PR #9, opened 2018-05-11
update pyre check to return nonzero exit code when errors are found

BenDoan:
I didn't see an obvious way of using `pyre check` with scripts or a CI system. This change updates `pyre check` to return a nonzero exit code if it sees any errors. This exit code can then be used to fail the build if errors are seen.


scripts/commands/initialize.py
```diff
@@ -14,7 +14,9 @@
     CONFIGURATION_FILE,
     EnvironmentException,
     find_typeshed,
-    log)
+    log
```
holmboe: Comma

BenDoan: Thanks, fixed.

BenDoan: @dark I couldn't figure out how to run the python tests. Are there docs anywhere that talk about that?

dkgi: We run the tests internally with buck. I suspect with the current setup this wont work on github. As far as I know the integration we rely on is not quite open sourced yet I'll defer to @ambv on that one.

dkgi:
I'm somewhat hesitant on this PR because I see how this makes sense but we also rely heavily on the current behavior. Is there a reason you're not using the results of the check to control whatever your CI system is going to do?

E.g. I could imagine it'd be fairly straightforward to wrap this in a script that reads out the JSON output (run `pyre --output=json check`) and count the number of results you get, fail if they're non-zero.

BenDoan:
Personally speaking, I would say I expect test and build tools to output exit codes when they fail. 

My current use case is just trying out pyre on a side project that uses Travis CI. Travis seems to solely [rely on the exit codes of commands to fail builds](https://docs.travis-ci.com/user/customizing-the-build#Breaking-the-Build). This been my experience with other CI tools as well.

I don't think it's *that* unreasonable not to implement this; it's not much work to add a wrapper script to a project. I do think, though, that it's generally expected for a tool like this to return a nonzero exit code when it "fails".

Regardless, thanks for the comments. I'm really digging pyre. Typed python is awesome, and pyre makes it super easy to use.

dark:
Assigning to myself to assess impact on current workflows.

scripts/__init__.py
```diff
@@ -28,6 +28,7 @@
 
 SUCCESS = 0
 FAILURE = 1
+HAS_ERROR = 2
```
dark: I would have `grep` semantics for the return value here: i.e., 0 for success, 1 for errors, and 2 for failures.

scripts/commands/incremental.py
```diff
@@ -83,3 +84,5 @@ def _run(self) -> None:
                 self.run()
             else:
                 exit(-1)
+
+        return SUCCESS
```
dark: This method should return `int` now.

scripts/commands/kill.py
```diff
@@ -26,7 +29,7 @@ def __init__(self, arguments, configuration, source_directory) -> None:
         super(Kill, self).__init__(arguments, configuration, source_directory)
         self.with_fire = arguments.with_fire
 
-    def _run(self) -> None:
+    def _run(self) -> int:
         if self.with_fire:
             self._kill_all_processes()
             return
```
dark: Need to return a value.

scripts/commands/start.py
```diff
@@ -25,7 +28,7 @@ def __init__(self, arguments, configuration, source_directory) -> None:
         self._no_watchman = arguments.no_watchman
         self._number_of_workers = configuration.number_of_workers
 
-    def _run(self) -> None:
+    def _run(self) -> int:
```
dark: There are a few `return` statements to be fixed in this method now.

dark:
Thanks for the update!
I have published a commit into the internal systems to accomodate this change, but they are still updating. We will need to wait a few days for that to happen.
I will review these changes again in the meantime.

scripts/commands/start.py
```diff
@@ -7,7 +7,11 @@
 import os
 
 from .. import filesystem
-from .command import Command
+from .command import (
+    Command,
+    FAILURE,
+    SUCCESS,
```
dark: I think the two imports above are in the wrong place - we should import from `..` instead.

dark:
Hi @BenDoan, our internal systems have been updated. We're ready to accept this contribution once the unit tests are fixed. Thanks!

BenDoan:
@dark Thanks for accommodating this PR! The tests should be passing now.

client/commands/check.py
```diff
@@ -26,7 +30,12 @@ def _flags(self):
             flags.extend(['-log-identifier', self._log_identifier])
         return flags
 
-    def _run(self, retries: int = 1) -> None:
+    def _run(self, retries: int = 1) -> int:
         result = self._call_client(command=self.NAME, flags=self._flags())
         errors = self._get_errors(result)
         self._print(errors)
+
+        if errors:
```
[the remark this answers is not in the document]
BenDoan: Changed

client/commands/start.py
```diff
@@ -25,7 +30,7 @@ def __init__(self, arguments, configuration, source_directory) -> None:
         self._no_watchman = arguments.no_watchman
         self._number_of_workers = configuration.number_of_workers
 
-    def _run(self) -> None:
+    def _run(self) -> int:
```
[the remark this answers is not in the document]
BenDoan: Fixed

client/commands/command.py
```diff
@@ -69,11 +69,11 @@ def __init__(
...
         self._current_directory = arguments.current_directory
 
-    def _run(self) -> None:
+    def _run(self) -> int:
         pass
 
-    def run(self) -> None:
-        self._run()
+    def run(self) -> int:
+        return self._run()
```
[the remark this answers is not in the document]
BenDoan: I personally like returning a status a little more. I think it's better to be explicit that _run() can return status and to show what that status is in each command.

I'm totally fine with the method you listed, though, if you or anyone else thinks that's the better way.

client/__init__.py
```diff
@@ -27,7 +27,8 @@
 
 
 SUCCESS = 0
-FAILURE = 1
+HAS_ERROR = 1
```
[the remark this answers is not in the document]
BenDoan: I agree, changed

client/commands/command.py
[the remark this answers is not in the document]
dark: I think that an explicit return type makes the intent much more evident. When debugging or understanding the code, it's much easier to track `return` statements rather than calls to other functions.

dark:
@BenDoan I have just realized that another piece of internal infrastructure needs an update. I will need to fix it before we can merge this change - I will update this PR when done.

We have also recently migrated our code to be auto-formatted through https://github.com/ambv/black. That caused a few rebase conflicts - could you please look at them? Thanks.

dark:
@BenDoan I have fixed the internal infra (again :) ).
We are now ready to accept the diff, once rebased.

Thanks again!

BenDoan:
@dark Just pushed an update. Thanks for all the help getting this feature through!

PR #22, opened 2018-05-12
Fix return type check for generators

### LINKED ISSUE #16: False positive error on `return` in generator
If you use `return` in a generator, an incorrect TypeError is generated:
```
$ cat srcdir/genreturn.py 
from typing import Iterable

def gen(x: int) -> Iterable[int]:
    for i in range(x):
        if i > 10:
            return
        else:
            yield i
$ pyre --source-directory srcdir/ check
 ƛ Setting up a .pyre_configuration file may reduce overhead.
 ƛ Found 1 type error!
srcdir/genreturn.py:6:12 Incompatible return type [7]: Expected `Iterable[int]` but got `None`.
$ python --version
Python 3.7.0b4
$ pip freeze | grep pyre  # related minor bug: "pyre --version" throws a TypeError
pyre-check==0.0.5
```

JelleZijlstra:
Fixes #16.

### DIFF
--- a/ast/astVisit.ml
+++ b/ast/astVisit.ml
@@ -394,7 +394,7 @@ let collect_accesses_in_position statement { Location.line; column } =
   Collector.collect (Source.create [statement])
 
 
-let contains_yield statements =
+let contains_yield {Statement.Define.body; _} =
   let module YieldVisitor = struct
     type t = bool
 
@@ -412,4 +412,4 @@ let contains_yield statements =
       | (false, _) -> false
   end in
   let module YieldVisit = Make(YieldVisitor) in
-  YieldVisit.visit false (Source.create statements)
+  YieldVisit.visit false (Source.create body)
--- a/ast/astVisit.ml
+++ b/ast/astVisit.ml
@@ -392,3 +392,24 @@ let collect_accesses_in_position statement { Location.line; column } =
         | _ -> None
     end) in
   Collector.collect (Source.create [statement])
+
+
+let contains_yield statements =
+  let module YieldVisitor = struct
+    type t = bool
+
+    let expression result expression =
+      match (result, expression) with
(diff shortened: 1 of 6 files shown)


ast/astVisit.ml
```diff
@@ -392,3 +392,24 @@ let collect_accesses_in_position statement { Location.line; column } =
...
+      | (false, {Node.value = Expression.Yield _; _}) -> true
+      | (false, _) -> false
+
+    let statement result statement =
+      match (result, statement) with
+      | (true, _) -> true
+      | (false, {Node.value = Statement.Yield _; _}) -> true
+      | (false, {Node.value = Statement.YieldFrom _; _}) -> true
+      | (false, _) -> false
+  end in
```
dkgi: We care about consistency in the code so here go a few style-nits. As a rule of thumb it's usually a good idea to be consistent with the surrounding code.

- I'd inline the module as above
- `in`s get their own separate line unless the statement fits on a single line (the one in the previous function is also inconsistent with the rest of our code, whoops ;) )
- space after `{` and before `}`
- The parentheses around the tuples are not necessary.

ast/astVisit.ml
```diff
@@ -392,3 +392,24 @@ let collect_accesses_in_position statement { Location.line; column } =
         | _ -> None
     end) in
   Collector.collect (Source.create [statement])
+
+
+let contains_yield {Statement.Define.body; _} =
```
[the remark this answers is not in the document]
JelleZijlstra: I tried that, but I couldn't get it too work because analysisTypeCheck.ml already uses the Annotation name for the file analysisAnnotation.ml. Do you mind if I put the function there instead?

ast/astVisit.ml
JelleZijlstra: Done.

PR #23, opened 2018-05-12
Add pipsi install instructions

tedmiston:
Hey all - in my experience, installing dev tooling like linters or type checkers like pyre system-wide via [pipsi](https://github.com/mitsuhiko/pipsi) vs once per project virtual environment has made it easier to always stay on the latest release everywhere.  I've added pipsi install instructions for others interested in a similar setup.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -24,6 +24,13 @@ pyre --source-directory path/to/project/root
 ```
 if you'd like to specify the path instead.
 
+Alternatively, install pyre via
+[pipsi](https://github.com/mitsuhiko/pipsi) to have a single
+system-wide install instead of one per virtual env:
+```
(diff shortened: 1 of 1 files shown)


dkgi:
@tedmiston: thanks for the contribution :) I'm not super familiar with the python ecosystem. @ambv: can you take a look a this?

tedmiston:
Okay, no worries, thank you for reviewing.  It is a bit of a letdown as a contributor.

`pipsi` is a small wrapper (<500 LoC) built for this use case.  Its core author @mitsuhiko is well known in Python as the creator of Flask.

The rationale is clear but somewhat unsatisfying given that the project documents or includes other third party tools similarly like VSCode, Atom, Django, and IPython.

PR #30, opened 2018-05-12
[setup] Fix error checking in setup.sh script

mzlee:
`test $var = 1` when `$var` is empty prints an error

    ./scripts/setup.sh: line 110: test: =: unary operator expected

With the variable quoted, it works as expected. (Tested on MacOS 10.12 bash v4.4)


dkgi:
Good catch :) Thanks!

PR #86, opened 2018-08-20
collect missing call edges

rvantonder:
A call signature (annotated access) is not created for some calls. When that is the case, record the call edge for a `Call` as revealed by the AST.

### DIFF
--- a/analysis/callGraph.ml
+++ b/analysis/callGraph.ml
@@ -38,21 +38,21 @@ let create ~environment ~source =
           |> Access.Map.of_tree
         in
         let resolution = Environment.resolution environment ~annotations () in
-        let fold_accesses call_graph { Node.value = access; _ } =
-          let add_call_edge call_graph ~resolution:_ ~resolved:_ ~element =
-            let open Annotated.Access in
-            let add_call_edge call_graph caller callee =
-              Log.log
-                ~section:`CallGraph
-                "Adding call edge %a -> %a"
-                Access.pp caller
-                Access.pp callee;
-              let update_callees = function
-                | Some callees -> callee :: callees
-                | None -> [callee]
-              in
-              Access.Map.update call_graph caller ~f:update_callees
+        let process_access call_graph access =
+          let add_call_edge call_graph caller callee =
+            Log.log
+              ~section:`CallGraph
+              "Adding call edge %a -> %a"
+              Access.pp caller
+              Access.pp callee;
+            let update_callees = function
+              | Some callees -> Access.Set.add callees callee
+              | None -> Access.Set.singleton callee
             in
+            Access.Map.update call_graph caller ~f:update_callees
+          in
+          let resolve_access _ ~resolution:_ ~resolved:_ ~element =
+            let open Annotated.Access in
             match element with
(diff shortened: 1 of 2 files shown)


analysis/callGraph.ml
```diff
@@ -65,22 +65,33 @@ let create ~environment ~source =
...
               } ->
-                add_call_edge call_graph caller callee
+                Some callee
             | _ ->
-                call_graph
+                None
           in
           Annotated.Access.create access
-          |> Annotated.Access.fold ~resolution ~initial:call_graph ~f:add_call_edge
+          |> Annotated.Access.fold ~resolution ~initial:None ~f:resolve_access
```
rvantonder: @dkgi, basically, I noticed that `Annotated.Access.fold` doesn't create a `Signature` for some types of calls (see test). Not sure if this is "as intended" or if there is a missing case that should be added in `Access.fold`. For now, the logic below will catch missing edges.

analysis/callGraph.ml
```diff
@@ -65,22 +65,33 @@ let create ~environment ~source =
...
+          |> function
+          | Some callee -> add_call_edge call_graph caller callee
+          | None ->
```
rvantonder: Basically, if we didn't resolve a call signature, check if the access is a `Access.Call` and then record it. @fahndrich this solves the missing edges issue and possibly fixes the low fixpoint iteration issue.

analysis/callGraph.ml
```diff
@@ -38,21 +38,21 @@ let create ~environment ~source =
...
+        let process_access call_graph access =
+          let add_call_edge call_graph caller callee =
+            Log.log
+              ~section:`CallGraph
+              "Adding call edge %a -> %a"
+              Access.pp caller
+              Access.pp callee;
+            let update_callees = function
+              | Some callees -> Access.Set.add callees callee
+              | None -> Access.Set.add Access.Set.empty callee
```
sinancepel: Just `Access.Set.singleton callee`?

analysis/callGraph.ml
dkgi: Not sure I see what you mean in the tests. Are we missing those just in conditional tests? If so, that sounds wrong.

Sounds like this should be fixed in the access fold. Let me know how you want to proceed. Fine with landing this and adding a TODO.

rvantonder: No it's not specific to conditional test (so I've removed that). It's just a minimal example of where call signatures are not created. 

> Sounds like this should be fixed in the access fold.

Possibly right? As a very simple example:

```python
def foo():
    bar()
```

When I fold over the `bar` access and print each element (there's only one), I see the resolution is only:

```
DUMP element: (AnnotatedAccess.Element.Attribute
   { AnnotatedAccess.Element.attribute = bar;
     origin = (AnnotatedAccess.Element.Module ); defined = false })
```

> Let me know how you want to proceed. Fine with landing this and adding a TODO.

Think this is fine for now until you have a chance to look into whether this is intended. Either way, the tests are good. Apparently you can edit my PR (e.g., add a TODO with task number), but I have put a placeholder TODO anyway.

PR #91, opened 2018-08-25
Few line changes to make install slightly smoother

KevinHock:
Add 'install coreutils' message if greadlink not installed
Set MACOSX_DEPLOYMENT_TARGET based off sw_vers (only resulted in warnings, not errors.)

Let me know if instead or in addition to the `coreutils` check, you would like me to mention it and [`watchman`](https://facebook.github.io/watchman/docs/install.html#buildinstall) in the [Building from Source](https://pyre-check.org/docs/installation.html#building-from-source) instructions.

Side note: Looking forward to making **many** more PRs over the next few years 😉 Pyre seems awesome

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -16,9 +16,22 @@ die() {
 DEVELOPMENT_COMPILER="4.06.0"
 RELEASE_COMPILER="4.06.0+flambda"
 
+version_less_than() {
+  if [ "$1" = "$2" ]; then
+    return 1
+  else
+    if ! [ -x "$(command -v gsort)" ]; then
+      die 'Error: gsort is not installed. Please install coreutils.'
+    fi
+    [ "$1" = "`echo -e "$1\n$2" | gsort -V | head -n1`" ]
+  fi
+}
+
 # Compatibility settings with MacOS.
 if [[ "${MACHTYPE}" = *apple* ]]; then
   export MACOSX_DEPLOYMENT_TARGET="$(sw_vers -productVersion | cut -d '.' -f 1,2)"
+  version_less_than $MACOSX_DEPLOYMENT_TARGET 10.11 \
+    && die 'Error: The minimum OSX version supported is 10.11, please update.'
 fi
 
 # Switch to pyre directory.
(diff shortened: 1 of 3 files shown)


sinancepel:
Hey Kevin, this looks great :) Excited that you feel that way; we're always happy to review contributions. I'm pulling this into our internal build system to ensure all tests pass before closing.

scripts/setup.sh
```diff
@@ -18,7 +18,7 @@ RELEASE_COMPILER="4.06.0+flambda"
 
 # Compatibility settings with MacOS.
 if [[ "${MACHTYPE}" = *apple* ]]; then
-  export MACOSX_DEPLOYMENT_TARGET=10.11
+  export MACOSX_DEPLOYMENT_TARGET="$(sw_vers -productVersion | cut -d '.' -f 1,2)"
```
dark: This hardcoded value is actually intended, it represents the minimum version of OSX supported by Pyre. Replacing it with the `sw_vers` command would make it dynamic instead, and always equivalent to the current version of OSX (hence, possibly wrong).

KevinHock: ++, Thanks for catching, will change 👍

dark:
Sorry, I wasn't too clear earlier.
I meant that the the `10.11` value is supposed to be exported statically, because it is used by the toolchain to generate code supporting versions as old as `10.11`, even if the current OSX version is greater than that. Extracting the value from `sw_vers` would simply make the compiler target the current OSX version, hence making `MACOSX_DEPLOYMENT_TARGET` redundant.

At a glance, the original code in `setup.sh` should stay as-is, otherwise we would not support `10.11` anymore.
However, you also mentioned that you are getting errors while compiling?

KevinHock:
My bad 🤦‍♂️, sorry about that.

re: `Set MACOSX_DEPLOYMENT_TARGET based off sw_vers (only resulted in warnings, not errors.) (only resulted in warnings, not errors.)` it was just warnings.

dark:
No problem!

If the warnings are similar to "cannot build for target 10.11 because some libraries are built targeting 10.13" (I don't remember the exact wording ATM), it means that some other pre-installed library from OPAM was installed with a wrong/different `MACOSX_DEPLOYMENT_TARGET` outside of pyre setup. If you are not targeting old MacOS installs, those warnings are also harmless.

PR #108, opened 2018-10-03
Update installation.md typo

stvreumi:
change `pip36` to `pip3 in the `installation.md`.


mrkmndz:
pip36 is intended, as that ensures we're using python > 3.5, which is what pyre supports.  Thanks!

PR #113, opened 2018-10-18
Ignore Error Types

pwgraham91:
This commit adds support for ignoring error types. One can ignore error types by error type code in the command line or in the configuration file.

### DIFF
--- a/client/commands/reporting.py
+++ b/client/commands/reporting.py
@@ -22,6 +22,7 @@ class Reporting(Command):
         self._verbose = arguments.verbose
         self._output = arguments.output
         self._do_not_check_paths = configuration.do_not_check
+        self._ignore_error_types = configuration.ignore_error_types
         self._discovered_analysis_directories = [self._local_root]
         self._local_configuration = arguments.local_configuration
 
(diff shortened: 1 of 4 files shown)


dark:
I think @dkgi would be the best to review this code - he has more ideas about our plan to extend the configurability of pyre via options.

dkgi:
Don't have strong opinions here except that I would like to avoid configurability-hell. @mrkmndz and @shannonzhu, thoughts?

pwgraham91:
FYI: I updated the branch to fix a merge conflict in configuration_test.py

dark:
Thanks, I will update the pull request internally.

Speaking of the solution and problem it's solving, I think this makes sense: some developers might want to disable some error classes for errors they do not care about.
I have not reviewed this code yet though - still waiting for feedback from the team.

dark:
Hi @pwgraham91, sorry for the huge delay in the review.
The team could not find a consensus on whether to merge this feature, and the argument was that we are trying to maintain a consistent experience across all users of Pyre.
Unfortunately we will not be able to accept this pull request at this time, but thanks for submitting the patch!

Issue #117, opened 2018-10-28
documentation request: how to follow imports within a virtualenv


SKalt:
### Issue Scope
1. provide an example in the guided tour on following imports in a virtualenv
2. clarify CLI argument documentation to show how to pass multiple directories to `--search-path`

### The Story
I wanted to write an example of checking pip-installed imports using pyre.  After reading through the documentation, I found that vanilla `source activate my-venv; pyre --source-directory to/check check` was not detecting packages I had `pip install`ed in my virtualenv.  Reading #64, I tried passing in a unix-style path, separated by `:`s. This failed with a semi-informative error message, `pyre_check.exceptions.EnvironmentException: colon/separated:list/of:paths/to/search is not a valid directory.`  I found passing in multiple `--search-paths` worked: 
```{sh}
get_path_to_check() { 
  python <<eof
import sys
from os.path import isdir
result = ' '.join(
    f'--search-path {i}' for i in sys.path
    if isdir(i) and 'site-packages' in i
)
print(result)
eof
}
pyre $(get_search_path) --source-directory to/validate check
```
This solution isn't 100% intuitive.  The guided tour should include an example like this of working in a virtualenv to make pyre easier to adopt. Updating the `--search-path SEARCH_PATH` documentation from
```
Additional directories with modules and stubs to include in type environment
```
to something like
```
Adds an additional directory with modules and stubs to include in type
environment each time it is passed
```
would also clarify pyre usage.

dark:
Thanks for the report, would you like to provide pull requests to clarify the usage of `--search-path` and how that can be used in a virtualenv?

SKalt:
Sure thing! Full disclosure: I'm only familiar with `pipenv` and `conda` envs.  I'll see if I can set up minimal gists for `pipenv` and `virtualenv`; I've got one for `conda`. 

Aside: looking through the code, I noticed [at ./client/configuration.py#Configuration.__init__](https://github.com/facebook/pyre-check/blob/128abb431a7fb038b629b0f086f4954da864a6a0/client/configuration.py#L91) it looks like you could ingest the `sys.path` that `pyre` is running in. Do you think that would make working with python envs simpler or introduce gotchas?

dark:
Sorry for the late reply, but this dropped from my radar somehow. I am looking at the PR, I have minor comments on that one.

[...]

PR #120, opened 2018-11-09
make --search-path documentation more explicit

SKalt:
This change relates to #117. I am to make it more explicit that `--search-path` is passed multiple times, rather than a single time with a delimited list of directories (e.g. `path/to:your/pyton:modules/and/stubs`)

### DIFF
--- a/client/pyre.py
+++ b/client/pyre.py
@@ -144,8 +144,8 @@ def main() -> int:
         action="append",
         default=[],
         type=readable_directory,
-        help="Additional directories with modules and stubs "
-        "to include in type environment",
+        help="adds an additional directory of modules and stubs to include in type"
+        " environment each time it is passed",
     )
     parser.add_argument(
         "--preserve-pythonpath",


client/pyre.py
```diff
@@ -144,8 +144,8 @@ def writable_path(path: str) -> str:
         action="append",
         default=[],
         type=readable_directory,
-        help="Additional directories with modules and stubs "
-        "to include in type environment",
+        help="adds an additional directory of modules and stubs to include in type"
+        " environment each time it is passed",
```
dark: I am essentially bikeshedding here, but I'd like to keep this consistent with the surrounding options.

I would prefer something like "Add an additional directory of modules and stubs to include in type environment":
* Starts with uppercase
* Uses imperative form rather than the present tense (I know, "Preserves the value of the PYTHONPATH..." just below is wrong)
* Avoids redundancy with the "each time it is passed" part

SKalt: Sure thing.

PR #123, opened 2018-11-22
Use virtualenv pythonpath

### LINKED ISSUE #121: Working in pipenv?
I encountered the problem described here: https://github.com/facebook/pyre-check/issues/64

I found that it's fairly easy to fix:

```
$ pyre --search-path=$(pipenv --venv)/lib/python3.7/site-packages check
```

I'm guessing this is something that pyre might be able to do automatically/transparently. Is this something I could set in the configuration (without having to commit the absolute path to my virtualenv, which is specific to my environment)?

SKalt:
@dark this is re: the comment you made in #117 about extending `--preserve-pythonpath`. When it's used, I include `sys.path` in the paths searched.  Please check the test output of  to verify the paths are as expected. With `PYTHONPATH=additional/:directories/`, it sets 
```python
>>> configuration.search_path
[
  'additional/',
  'directories/',
  # vvv the contents of sys.path that are directories
  '',
   '/your/env/anaconda3/lib/python3.6',
   '/your/env/anaconda3/lib/python3.6/lib-dynload',
   '/your/env/anaconda3/lib/python3.6/site-packages',
   # ^^^
   'command/',
   'line/',
   'json/',
   'file/',
]
```

Assuming my understanding is correct and python searches `$PYTHONPATH` and then `sys.path` for packages, then this should fix #121 as well.


client/configuration.py
```diff
@@ -88,9 +89,9 @@ def __init__(
 
         # Handle search path from multiple sources
         self.search_path = []
-        pythonpath = os.getenv("PYTHONPATH")
+        pythonpath = os.getenv("PYTHONPATH", default='')
         if preserve_pythonpath and pythonpath:
```
dark: This line will now fail to execute the import logic if `--preserve-pythonpath` is used, but `PYTHONPATH` itself is empty. The logic needs to be changed to address this case, but beware below that:
```
>>> ''.split(":")
['']
```

SKalt: Logic changed.  I'm satisfied the `if os.path.isdir(path)` section will filter out bad strings from missing `$PYTHONPATH`s and malformed `sys.path`s.

dark: Sort of - a warning will still be emitted below in that case.
So, in case of an empty string, a confusing warning would be printed. I think we should avoid doing that.

Now, I just noticed that the warning message mentions `PYTHONPATH` explicitly. For the same reasons of clarity, I think it's worth splitting the loop over `PYTHONPATH` from the one over `sys.path`, to have two different, but more precise error messages.

dark:
Left one comment about the logic.
Could you please amend the documentation as well, to mention the new behaviour? Thanks.

docs/configuration.md
```diff
@@ -102,7 +102,8 @@ optional arguments:
                         include in the type environment
   --preserve-pythonpath
                         Preserve the value of the PYTHONPATH environment
-                        variable
+                        variable and inheret the current python environment's
```
dark: Same spelling fix needed here.

client/configuration.py
```diff
@@ -97,16 +98,20 @@ def __init__(
...
+                        self.search_path.append(path)
+                    else:
+                        LOG.warning(
+                            "'{}' is not a valid directory, dropping it "
+                            "from PYTHONPATH".format(path)
+                        )
+            sys_path = [path for path in sys.path if os.path.isdir(path)]
+            self.search_path.extend(sys_path)
+            # sys.path often includes '' and a zipped python version, so
+            # we don't log warnings for non-dir entries
```
dark: We usually leave comments above the code they comment on.

client/configuration.py
```diff
@@ -131,17 +132,24 @@ def __init__(
...
-        if preserve_pythonpath and pythonpath:
-            for path in pythonpath.split(":"):
-                if os.path.isdir(path):
-                    self._search_path.append(SearchPathElement(path))
-                else:
-                    LOG.warning(
-                        "`{}` is not a valid directory, dropping it "
-                        "from PYTHONPATH".format(path)
-                    )
+        self.search_path = []
```
dark: Something went wrong with the rebase I think, you accidentally renamed `self._search_path` to `self.search_path` in a couple of places.

docs/configuration.md
dark: Spelling error still present?

dark:
I took the liberty to fix those issues myself :) (and running the `black` linter on the PR as well)

PR #127, opened 2018-12-19
[RFC] Fix f-string analysis

### LINKED ISSUE #97: Support double curly braces in f-strings
The following simple code:

`print(f'{{x}}')`

produces an error about undefined name:

> Undefined name [18]: Global name `x` is undefined.

(pyre version 0.0.10)

jas14:
Format strings allow escape sequences for curly braces, e.g. `f"{{x}}"` evaluates to the literal `"{x}"`. Fixes #97 .

NOTE: this is actually pretty bad - as you can see, it doesn't support nested expressions. `{{{x}}}` should be equivalent to `"{" + x + "}"`, but we're missing it altogether here.

I think the f-string analysis bit needs to be a state machine. I can do that, but in the meantime, RFC on conventions please!


jas14:
abandoning in favor of #128

PR #150, opened 2019-04-18
Don't block for so long in persistent client

rvantonder:
Select here blocks for up to 1 second. This can lead to behavior where sending two LSP messages in quick succession appear to get buffered, but only one message is read/processed at a time (until, e.g., more input becomes available). Clients calling persistent may wait for the requests to be fulfilled (and therefore wait forever) if not all the input received is processed by persistent.

This change lets the persistent process immediately on available input. No easy way to test this with OUnit, so I'm punting :(

### DIFF
--- a/command/persistentCommand.ml
+++ b/command/persistentCommand.ml
@@ -10,8 +10,6 @@ open Pyre
 open Network
 open Server
 
-module Time = Core_kernel.Time_ns.Span
-
 
 type reason = string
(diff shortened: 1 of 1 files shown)


sinancepel:
Seems reasonable. We use an explicit timeout for the server handshake since we want to give the server a time to initialize, but that shouldn't be necessary once we've connected.

PR #156, opened 2019-04-27
Add sqlite3 lib ccflag for hack_parallel.heap

jas14:
The hack_parallel.heap library relies on sqlite3 functions, so `make` at the project root fails without this. Logs [here](https://pastebin.com/YSdTM07H).

Also, Q: where else should the project's C deps be listed?


sinancepel:
@jas14, any chance you could try adding the dependency of `conf-sqlite3` on the opam file?

jas14:
@sinancepel sure, I'll do that when I get back to a computer in a couple weeks. feel free to do so beforehand if you haven't already!

PR #166, opened 2019-05-03
Wrap config paths in os.path.expanduser

### LINKED ISSUE #165: Committing config - avoiding hardcoded paths with user names
I'm looking to use Pyre in a setting where developers work on the project on local workstations under different users, which means that paths to typeshed such as `"/home/hawkins/.local/lib/pyre_check/typeshed/"` are problematic for other developers, because they now have to replace `hawkins` with whatever their username may be.

I've tried replacing this fixed path with `"~/.local/lib/pyre_check/typeshed"`. `/home/$USER/.local/lib/pyre_check/typeshed/"`, but both fail saying they're invalid paths.

i.e.: 

```
 ƛ ~/.local/lib/pyre_check/typeshed/ is not a valid directory.
 ƛ $HOME/.local/lib/pyre_check/typeshed/ is not a valid directory.
 ƛ /home/$USER/.local/lib/pyre_check/typeshed/ is not a valid directory.
```

What's the idiomatic way to set up paths in a pyre_config to be committed to a project? Or is there some other pattern we should be following (unique non-committed pyre config files in our home dir, etc)?

hawkins:
Intended to fix #165 

---

The idea was to cover every spot a path can be supplied via config file, though some of these places may not make much sense. For instance, I'm not familiar enough with Pyre to know if it does in both root and subdirectory of a `SearchPathElement` dictionary.

I haven't used pyre a ton yet, but I pretty quickly found a need for this in at least `typeshed`, `source_directories`, and `binary`. Open to any sorts of feedback on this!

---

This is opened as a draft because I would like to be open to feedback at any stage since I'm so new to Pyre, so as such there's still a little bit to do before this is ready for any final reviews/merging.

Still to-do:

- [x] Discuss whether it makes sense on each of these fields or anywhere else in Pyre
- [x] Patch `os.path.expanduser` where necessary to pass preexisting tests
- [x] Add new tests to each relevant field
- [x] Update documentation
- [x] Lint

### DIFF
--- a/client/configuration.py
+++ b/client/configuration.py
@@ -354,11 +354,16 @@ class Configuration:
                 configuration_directory = os.path.dirname(path)
                 if configuration_directory:
                     self.source_directories = [
+                        # TODO: Does it make sense to expanduser
+                        # when a configuration directory is present?
                         os.path.join(configuration_directory, os.path.expanduser(directory))
                         for directory in source_directories
                     ]
                 else:
-                    self.source_directories = source_directories
+                    self.source_directories = [
+                        os.path.expanduser(directory)
+                        for directory in source_directories
+                    ]
 
                 self.targets = configuration.consume(
                     "targets", default=[], current=self.targets, print_on_success=True
--- a/client/configuration.py
+++ b/client/configuration.py
@@ -39,7 +39,7 @@ class SearchPathElement:
                     "Search path elements must have `root` and `subdirectory` specified."
                 )
             self.root = os.path.abspath(os.path.expanduser(element["root"]))
-            self.subdirectory = os.path.expanduser(element["subdirectory"])
+            self.subdirectory = element["subdirectory"]
 
     def path(self) -> str:
         subdirectory = self.subdirectory
--- a/client/configuration.py
+++ b/client/configuration.py
@@ -31,15 +31,15 @@ class InvalidConfiguration(Exception):
 class SearchPathElement:
     def __init__(self, element: Union[Dict[str, str], str]) -> None:
         if isinstance(element, str):
-            self.root = os.path.abspath(element)
+            self.root = os.path.abspath(os.path.expanduser(element))
             self.subdirectory = None
         else:
             if "root" not in element or "subdirectory" not in element:
                 raise InvalidConfiguration(
                     "Search path elements must have `root` and `subdirectory` specified."
                 )
-            self.root = os.path.abspath(element["root"])
-            self.subdirectory = element["subdirectory"]
+            self.root = os.path.abspath(os.path.expanduser(element["root"]))
+            self.subdirectory = os.path.expanduser(element["subdirectory"])
 
     def path(self) -> str:
         subdirectory = self.subdirectory
@@ -354,7 +354,7 @@ class Configuration:
                 configuration_directory = os.path.dirname(path)
                 if configuration_directory:
                     self.source_directories = [
-                        os.path.join(configuration_directory, directory)
+                        os.path.join(configuration_directory, os.path.expanduser(directory))
                         for directory in source_directories
                     ]
                 else:
@@ -396,7 +396,7 @@ class Configuration:
 
                 binary = configuration.consume("binary", current=self._binary)
                 assert binary is None or isinstance(binary, str)
-                self._binary = binary
+                self._binary = os.path.expanduser(binary)
 
                 additional_search_path = configuration.consume(
                     "search_path", default=[]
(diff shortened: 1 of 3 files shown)


client/configuration.py
```diff
@@ -354,11 +354,16 @@ def _read(self, path: str) -> None:
                 configuration_directory = os.path.dirname(path)
                 if configuration_directory:
                     self.source_directories = [
-                        os.path.join(configuration_directory, directory)
+                        # TODO: Does it make sense to expanduser
+                        # when a configuration directory is present?
+                        os.path.join(configuration_directory, os.path.expanduser(directory))
```
sinancepel: Since the directory is forced to be a relative path here, it seems unlikely that expanding `~` would change anything here. As a side note, checking whether the dirname of the path is not None seems a bit finicky. I suppose the proper thing to do here would be to detect whether the source directory is an absolute path and avoid joining it with the configuration directory in the first place, but that's definitely outside the scope of this PR.

client/configuration.py
```diff
@@ -396,6 +401,8 @@ def _read(self, path: str) -> None:
 
                 binary = configuration.consume("binary", current=self._binary)
                 assert binary is None or isinstance(binary, str)
+                if isinstance(binary, str):
```
sinancepel: Nit: `binary is not None` reads a bit more cleanly to me :)

client/tests/configuration_test.py
```diff
@@ -181,6 +186,21 @@ def test_init(
         self.assertEqual(configuration.search_path, ["simple_string/"])
         self.assertEqual(configuration.file_hash, "HASH")
 
+        expanduser.side_effect = [
+            "/home/user/simple_string_searchpath",
+            "/home/user/simple_string_searchpath"
+        ]
+        json_load.side_effect = [
+            {
+                "search_path": "~/simple_string_searchpath",
```
sinancepel: `search_path` is meant to be a list of search path elements - could you rewrite the test in the form of
`"search_path": ["~/simple", {"root": "~/simple", "subdirectory": "subdir"}]`?

sinancepel:
Forgot to mention: The general implementation seems reasonable, I don't think you missed any fields here.

client/configuration.py
hawkins: Great point, that's a faster check too - thanks!

client/tests/configuration_test.py
hawkins: Yes, thanks - it's a very rough first pass at a test case, so I definitely goofed that. I'm still planning to add more tests for the other fields (`typeshed`, `binary`, etc) as well

client/configuration.py
hawkins: That's more or less what I was thinking, thanks. Also agree with the side note, but I'll avoid that in this PR for scope and test complexity while I'm still unfamiliar 🙂

client/configuration.py
```diff
@@ -31,15 +31,15 @@ class InvalidConfiguration(Exception):
...
             self.subdirectory = None
         else:
             if "root" not in element or "subdirectory" not in element:
                 raise InvalidConfiguration(
                     "Search path elements must have `root` and `subdirectory` specified."
                 )
-            self.root = os.path.abspath(element["root"])
-            self.subdirectory = element["subdirectory"]
+            self.root = os.path.abspath(os.path.expanduser(element["root"]))
+            self.subdirectory = os.path.expanduser(element["subdirectory"])
```
sinancepel: Do we need to expand here? subdirectory should always be a relative path

client/tests/configuration_test.py
```diff
@@ -54,6 +56,7 @@ def test_init(
             },
             {},
         ]
+        expanduser.side_effect = ["a"]
```
sinancepel: One neat trick with mock is that you can actually make side effects of mocked functions into lambdas. I have a suspicion this might save us a bit of code in this case :) Looks like we're consistently expanding "~" into "/home/user" in the below tests, so this code suggestion might save you the effort of writing all those side effects below.
```suggestion
        expanduser.side_effect = lambda path: path.replace("~", "/home/user")
```

client/configuration.py
hawkins: No, I wasn't 100% positive of that, but knowing that it's relative I don't believe we need it under `subdirectory`. I would say we do under `root` though

client/tests/configuration_test.py
hawkins: Very solid point, wish I'd recognized that sooner! Thank you! That's sufficient for all of the configuration testing 🎉

sinancepel:
@hawkins  could you run black + isort on this and add a line of documentation? Let's merge this :)

hawkins:
Almost seems silly to say, but I'm on the fence about this change even needing a reference in documentation.

It doesn't seem too fit anywhere super well, unless I added a section specific to user directory expansion on the configuration page to elaborate which fields have the capability and which do not, but that almost seems too in-the-weeds for what strikes me as a common-sense change. We expand anywhere it makes sense, and don't where it doesn't.

What would you think of not changing docs for this, @sinancepel ?

PR #171, opened 2019-05-13
Fix SAPP CI

jas14:
97b151c9399 moved sapp/ to tools/sapp, Travis config needs to be updated accordingly


dkgi:
Thanks for fixing this. I was pretty sure I would something but wasn't quite sure what.

PR #225, opened 2019-12-21
Unbreak sapp CI build

SamChou19815:
`requirements.txt` is removed in [this commit](https://github.com/facebook/pyre-check/commit/4288a9500753f016b0e36a9e69234d80841ec3aa)
The requirements are already automatically installed during setup. Therefore, we don't need to run `pip install -r requirements.txt`, and we can simply remove the line in CI config.


gbleaney:
Good catch. Thanks for the PR!

PR #231, opened 2020-01-24
Query min trace length - SAPP Update

fyunaz:
This PR update Static Analysis Post Processor (SAPP) to filter issues based on its `min_trace_length_to_sources` and `min_trace_length_to_sinks`. These capabilities help users to easily query and narrow down issues when doing the analysis.

Filter addition to issues command:
`minTraceLength_to_sources`: int or list[int]   -> min trace length to sources to filter on 
`minTraceLength_to_sinks`: int or list[int]   -> min trace length to sinks to filter on 
`minTraceLength_max_to_sources`: int  ->  maximum min trace length to sources to filter on 
`minTraceLength_max_to_sinks`: int   -> maximum min trace length to sinks to filter on 

Sample:
```
[ run 1 ]                                                                                                                                                                                                   
>>> issues(minTraceLength_to_sources=0)                                                                                                                                                                 
Issue 1
            Code: 5001
         Message: Possible shell injection Data from [UserSpecified] source(s) may reach [RemoteCodeExecution] sink(s)
        Callable: source.convert
         Sources: UserSpecified
           Sinks: RemoteCodeExecution
        Location: source.py:9|22|32
Min Trace Length: Source (0) | Sink (1)
Found 1 issues with run_id 1.

[ run 1 ]                                                                                                                                                                                                   
>>> issues(minTraceLength_max_to_sources=0,minTraceLength_max_to_sinks=0)                                                                                                                                   

Found 0 issues with run_id 1.
```

### DIFF
--- a/tools/sapp/sapp/interactive.py
+++ b/tools/sapp/sapp/interactive.py
@@ -445,10 +445,10 @@ details              show additional information about the current trace frame
         codes: Optional[Union[int, List[int]]] = None,
         callables: Optional[Union[str, List[str]]] = None,
         filenames: Optional[Union[str, List[str]]] = None,
-        minTraceLength_to_sources: Optional[Union[int, List[int]]] = None,
-        minTraceLength_to_sinks: Optional[Union[int, List[int]]] = None,
-        minTraceLength_max_to_sources: Optional[int] = None,
-        minTraceLength_max_to_sinks: Optional[int] = None,
+        exact_trace_length_to_sources: Optional[Union[int, List[int]]] = None,
+        exact_trace_length_to_sinks: Optional[Union[int, List[int]]] = None,
+        max_trace_length_to_sources: Optional[int] = None,
+        max_trace_length_to_sinks: Optional[int] = None,
     ):
         """Lists issues for the selected run.
 
@@ -457,13 +457,13 @@ details              show additional information about the current trace frame
             codes: int or list[int]        issue codes to filter on
             callables: str or list[str]    callables to filter on (supports wildcards)
             filenames: str or list[str]    filenames to filter on (supports wildcards)
-            minTraceLength_to_sources: 
+            exact_trace_length_to_sources: 
                 int or list[int]    min trace length to sources to filter on 
-            minTraceLength_to_sinks: 
+            exact_trace_length_to_sinks: 
                 int or list[int]    min trace length to sinks to filter on 
-            minTraceLength_max_to_sources: 
+            max_trace_length_to_sources: 
                 int    maximum min trace length to sources to filter on 
-            minTraceLength_max_to_sinks: 
+            max_trace_length_to_sinks: 
                 int    maximum min trace length to sinks to filter on 
 
         String filters support LIKE wildcards (%, _) from SQL:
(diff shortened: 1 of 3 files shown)


tools/sapp/sapp/interactive.py
```diff
@@ -445,6 +445,10 @@ def issues(
         codes: Optional[Union[int, List[int]]] = None,
         callables: Optional[Union[str, List[str]]] = None,
         filenames: Optional[Union[str, List[str]]] = None,
+        minTraceLength_to_sources: Optional[Union[int, List[int]]] = None,
```
sinancepel: I'm going to be an awful person, and suggest that we use a less precise name :) How would you feel about renaming `minTraceLength -> traceLength`? For the simple case (one path from source/to sink), it coincides with people's intuition, and the `minTraceLength` isn't an obvious concept unless you know how these tools work - so I think the term warrants a comment in docs either way.

The second suggestion I'd make is renaming these to be `exact` and `max` instead of `<empty string>` and `max`. I don't think it's currently clear that the `minTraceLength_to_sources` is an exact match, so might confuse the n00b :)

All in all, here's what the suggested names look like:

```
max_trace_length_to_sinks
exact_trace_length_to_sinks
max_trace_length_to_sources
exact_trace_length_to_sources
```

I'm not attached to these names at all, and you might have better ideas on naming these (after all, you're working with these daily :)) but hopefully the concern makes sense!

fyunaz: Yes, that all make sense. Totally agreed that the `max` in the min trace length seems a bit confusing :). I think those suggested names works for us. I'll update the code to match those parameters.

tools/sapp/sapp/tests/interactive_test.py
```diff
@@ -251,6 +257,119 @@ def testListIssuesFilterFilenames(self):
...
+
+        self.interactive.issues(max_trace_length_to_sinks=2)
+        output = self.stdout.getvalue().strip()
+        self.assertIn("Issue 1", output)
+        self.assertIn("Issue 2", output)
+        self.assertNotIn("Issue 3", output)
+        self._clear_stdout()
+
+        self.interactive.issues(
+            max_trace_length_to_sources=1, max_trace_length_to_sinks=1,
```
gbleaney: `black` [strikes again](https://our.internmc.facebook.com/intern/diff/D19503869/). Looks like it doesn't like the trailing comas

tools/sapp/sapp/interactive.py
```diff
@@ -453,6 +457,14 @@ def issues(
             codes: int or list[int]        issue codes to filter on
             callables: str or list[str]    callables to filter on (supports wildcards)
             filenames: str or list[str]    filenames to filter on (supports wildcards)
+            exact_trace_length_to_sources: 
+                int or list[int]    min trace length to sources to filter on
```
gbleaney: Might be worth updating the description to mirror the naming. Something like:
```
exact values for min trace length to {sources,sinks} to filter on 
```
and 
```
maximum value for min trace length to {sources,sinks} to filter on 
```

sinancepel:
It seems like the interactive test still has references to `min_trace_length_to_sources` - is that pre-existing?

fyunaz:
@sinancepel Can you please point me which code are you referring to? I think those are part of the `IssueInstance` model.

gbleaney:
@sinancepel, @fyunaz is correct, that code is part of `IssueInstance` and has been around for a while:
https://github.com/facebook/pyre-check/blame/77e80ddac0709410d6b1f383eb01d4775d8ff5ea/tools/sapp/sapp/models.py#L734

gbleaney:
Kicked off the land process. Sorry for the delay @fyunaz, and thanks again for the PR!

PR #234, opened 2020-02-05
Sapp: show features/breadcrumbs

fyunaz:
This PR update Static Analysis Post Processor (SAPP) to show issue's features/breadcrumbs. It will print the `Features:` of an issue when calling `issues/issue` command. These information is helpful for users when doing the analysis.

Sample:
```
>>> issues                                                                                                                  
Issue 2
            Code: 5001
         Message: Possible shell injection Data from [UserSpecified] source(s) may reach [RemoteCodeExecution] sink(s)
        Callable: source.convert
         Sources: UserSpecified
           Sinks: RemoteCodeExecution
        Features: always-via:input
                  always-via:obscure
        Location: source.py:9|22|32
Min Trace Length: Source (0) | Sink (1)
Found 1 issues with run_id 2.
```

### DIFF
--- a/tools/sapp/sapp/interactive.py
+++ b/tools/sapp/sapp/interactive.py
@@ -201,8 +201,10 @@ details              show additional information about the current trace frame
 
         self.sources: Set[str] = set()
         self.sinks: Set[str] = set()
+        self.features: Set[str] = set()
         self.sources_dict: Dict[int, str] = {}
         self.sinks_dict: Dict[int, str] = {}
+        self.features_dict: Dict[int, str] = {}
 
         # Tuples representing the trace of the current issue
         self.trace_tuples: List[TraceTuple] = []
@@ -223,6 +225,9 @@ details              show additional information about the current trace frame
 
             self.sources_dict = self._all_leaves_by_kind(session, SharedTextKind.SOURCE)
             self.sinks_dict = self._all_leaves_by_kind(session, SharedTextKind.SINK)
+            self.features_dict = self._all_leaves_by_kind(
+                session, SharedTextKind.FEATURE
+            )
 
         print("=" * len(self.welcome_message))
         print(self.welcome_message)
@@ -409,6 +414,10 @@ details              show additional information about the current trace frame
                 session, issue_instance_id, SharedTextKind.SINK
             )
 
+            self.features = self._get_leaves_issue_instance(
+                session, issue_instance_id, SharedTextKind.FEATURE
+            )
+
         self.current_issue_instance_id = int(selected_issue.id)
         self.current_frame_id = -1
         self.current_trace_frame_index = 1  # first one after the source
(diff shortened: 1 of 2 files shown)


tools/sapp/sapp/interactive.py
```diff
@@ -580,10 +589,17 @@ def issues(
...
+                )
+                for issue in issues
+            ]
             issue_strings = [
-                self._create_issue_output_string(issue, sources, sinks)
-                for issue, sources, sinks in zip(issues, sources_list, sinks_list)
+                self._create_issue_output_string(issue, sources, sinks, features)
+                for issue, sources, sinks, features in zip(
+                    issues, sources_list, sinks_list, features_list
+                )
```
gbleaney: Feel free to ignore, but this could be a good application of [`starmap`](https://docs.python.org/3.7/library/itertools.html#itertools.starmap)

tools/sapp/sapp/interactive.py
```diff
@@ -1648,17 +1673,22 @@ def _get_leaves_trace_frame(
...
         self, message_ids: List[int], kind: SharedTextKind
     ) -> Set[str]:
-        leaf_dict = (
-            self.sources_dict if kind == SharedTextKind.SOURCE else self.sinks_dict
-        )
+        if kind == SharedTextKind.SOURCE:
+            leaf_dict = self.sources_dict
+        elif kind == SharedTextKind.SINK:
+            leaf_dict = self.sinks_dict
+        else:
```
gbleaney: I have a minor preference for explicitly `elif`ing on `SharedTextKind.FEATURE` here and raising in the `else`, just as a sanity check. Feel free to ignore though, since you're already following the previously existing pattern of using the `else` to assume the expected `SharedTextKind`

fyunaz: Looking at the use of `raise` on the code, it's more tied to the `UserError` (used when a user make a mistake with their command). It looks like user don't have the ability to directly call this function and this mostly will be used internally inside the code. We can just use `elfi` and don't have an `else` or use switch here but I think error with stacktrace would be more helpful for future development instead of raise.

tools/sapp/sapp/interactive.py
fyunaz: It might be possible but it require more code changes. We need to construct a new iterable that consist of
 `issue, sources, sinks, features` with the same index element to do `starmap(self._create_issue_output_string, new_iterable)`. I think zip provides this and it's easier to read but let me know if this is not what you are thinking on this.

gbleaney: I was thinking of preserving the `zip`, and just removing the repetition of `issue, sources, sinks, features`:
```
starmap(self._create_issue_output_string, zip(issues, sources_list, sinks_list, features_list))
```
I'll yield the point that the original way is probably easier to read, and doesn't require people to look up some obscure `itertools` function :)

gbleaney:
I've kicked off the land process. Thanks for yet another handy improvement to SAPP! I really appreciate the work

PR #239, opened 2020-02-24
Small Pysa tutorial and pysa_precollectors.md edits

KevinHock:
:mortar_board: Small `pysa_tutorial/` and `docs/pysa_precollectors.md` edits

- Added missing TODO in `exercise4/generate_models.py`
- `are` -> `is` in `docs/pysa_precollectors.md`
- Small miscellaneous edits of `pysa_tutorial/`

### DIFF
--- a/pysa_tutorial/exercise3/README.md
+++ b/pysa_tutorial/exercise3/README.md
@@ -2,11 +2,11 @@
 ## Overview
 The purpose of this exercise is to learn how to remove false positive issues using _sanitizers_.
 
-None of the functions in `views.py` are vulnerable, but they will all show up in `pyre analyze` as false positives. The goal of this exercise is the make changes to `sources_sinks.pysa` and `views.py` in order to remove the false positives.
+None of the functions in `views.py` are vulnerable, but they will all show up in `pyre analyze` as false positives. The goal of this exercise is to make changes to `sources_sinks.pysa` and `views.py` in order to remove the false positives.
 
 ## What you need to know
 ### Pre-written Taint Annotations
-Pysa comes with pre-written sources, sinks, and rules for much of the Python standard library, and many open source libraries. Since the previous exercises already covered those concepts, from this exercise forward we will be relying on pre-written sources, sinks, and rules where we can. You can look at collection of pre-written `taint.config` and `.pysa` files in the `stubs/taint` folder in this repository. If you're curious about what changed between this and the previous exercise to allow us to take advantage of those pre-written files, compare the `.pyre_configuration` files.
+Pysa comes with pre-written sources, sinks, and rules for much of the Python standard library, and many open source libraries. Since the previous exercises already covered those concepts, from this exercise forward we will be relying on pre-written sources, sinks, and rules where we can. You can look at the collection of pre-written `taint.config` and `.pysa` files in the `stubs/taint` folder in this repository. If you're curious about what changed between this and the previous exercise to allow us to take advantage of those pre-written files, compare the `.pyre_configuration` files.
 
 ### Sanitizers
 _Sanitizers_ are defined in `.pysa` files, just like sources and sinks. An example is provided in `sanitizers.pysa` and you'll have to add more of your own.
--- a/pysa_tutorial/exercise4/generate_models.py
+++ b/pysa_tutorial/exercise4/generate_models.py
@@ -26,20 +26,19 @@ class Ignore:
 def main() -> None:
     # Here, specify all the generators that you might want to call.
     generators = {
-        # TODO
-        # "replace_me": generate_taint_models.SomeGenerator(
-        #     ...
-        # ),
+        "get_REST_api_sources": generate_taint_models.RESTApiSourceGenerator(
+            django_urls=view_generator.DjangoUrls(
+                urls_module="urls",
+                url_pattern_type=UrlPattern,
+                url_resolver_type=Ignore,
+            )
+        )
     }
     # The `run_generators` function will take care of parsing command-line arguments, as
     # well as executing the generators specified in `default_modes` unless you pass in a
     # specific set from the command line.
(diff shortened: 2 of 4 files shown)


pysa_tutorial/exercise3/README.md
```diff
@@ -2,11 +2,11 @@
...
-Pysa comes with pre-written sources, sinks, and rules for much of the Python standard library, and many open source libraries. Since the previous exercises already covered those concepts, from this exercise forward we will be relying on pre-written sources, sinks, and rules where we can. You can look at collection of pre-written `taint.config` and `.pysa` files in the `stubs/taint` folder in this repository. If you're curious about what changed between this and the previous exercise to allow us to take advantage of those pre-written files, compare the `.pyre_configuration` files.
+Pysa comes with pre-written sources, sinks, and rules for much of the Python standard library, and many open source libraries. Since the previous exercises already covered those concepts, from this exercise forward we will be relying on pre-written sources, sinks, and rules where we can. You can look at the collection of pre-written `taint.config` and `.pysa` files in the `stubs/taint` folder in this repository. If you're curious about what changed between this and the previous exercise to allow us to take advantage of those pre-written files, compare the `.pyre_configuration` files.
```
KevinHock: Difficult to see if the diff, just added `the` before collection.

KevinHock:
Really great tutorial!

I found it pretty easy and straightforward -- was able to run Pysa on a Pyramid app in no time! 💯 🥇Stubs were the only pain point, but I just copied them from the Django directory.

Is it possible to run it on e.g. git diffs, instead of however watchman invokes it?

For service to service calls, what do you do? I guess I can make `some call to Service A` a sink, write the output to a file, and then read that file inside the model generator for Service A. Mostly just curious; I'll leave that as low-priority and happily take any false-positives for the time being.

pysa_tutorial/exercise4/generate_models.py
```diff
@@ -26,19 +26,20 @@ class Ignore:
 def main() -> None:
     # Here, specify all the generators that you might want to call.
     generators = {
-        "get_REST_api_sources": generate_taint_models.RESTApiSourceGenerator(
```
gbleaney: I like the idea of providing people a template of how to call the model generators, since this will be the first time in the tutorial that they've encountered it. The generators are not well documented right now, so it's also nice to keep this as a working example to point people to.

What if we leave `get_REST_api_sources` fully implemented here, and remove this line from the instructions for `operate_on_twos`:
```
1. Replace the TODO comment sections in `generate_models.py` with the correct dynamic model generator to taint a Django-style API
```

The instructions for `operate_on_threes` can be the first time the reader has to configure a model generator on their own, and they can copy-paste these lines as a starting point

KevinHock: ++, makes sense to me, thanks!

KevinHock:
>Are you thinking of incremental runs for pysa (pyre analyze)?

Yup

> We don't support anything incremental with pysa right now unfortunately, so you have to re-analyze the whole application each time.

No worries, it is somewhat of a nice-to-have.

> Can you elaborate on this?

Sure, I made e.g. `pyramid.request.Request.swagger_data: TaintSource[UserSpecified] = ...` a sink, and then copied the `django/http/request.pyi` file, to a `pyramid` folder, added that one attribute (`swagger_data`) to the stub, and things worked perfectly 💯. I figure adding stubs for all sources/sinks/sanitizers/etc. will be a small one-time cost, and then everything else 'just works.' So it is really easy-to-use.

> Is there a way we can make them easier to use?

It's pretty great already 😁 but the main things I can think of are
- This may or may not make any sense, I should probably think about it more, but: similar to `If taint flows into a function Pysa doesn't have the source for, it will assume that the return type of that function has the same taint.`from [here](https://pyre-check.org/docs/pysa-basics.html#taint-analysis), if someone adds a source to the `sources_sinks.pysa` file, why does it need stubs? A function that Pysa doesn't have the source code for could return a dictionary, or a string, but the return value is still marked as tainted, (I think.) So maybe the same thing is possible with sources, not sure.
- After adding universal sanitizers, more sanitizers to the standard `stubs/` directory would be helpful. (Already on the roadmap.)

gbleaney:
> I figure adding stubs for all sources/sinks/sanitizers/etc. will be a small one-time cost, and then everything else 'just works.'

Yup. The ideal state is we'd have coverage for all the major Python frameworks, so starting up Pysa on a new app wouldn't require writing any models or stubs to get started. On that note, if you want to put any of those Pyramid stubs into a PR I'll happily merge it :)

> if someone adds a source to the `sources_sinks.pysa` file, why does it need stubs?

In the simple case for taint analysis, the stubs are kind of redundant, but they can be useful. For example, there are [a bunch of sub-types of HttpResponse](https://github.com/facebook/pyre-check/blob/master/stubs/django/http/response.pyi) in Django. If we didn't have stubs reflecting that inheritance relationship, [the single `HttpResponse` model](https://github.com/facebook/pyre-check/blob/master/stubs/taint/general_django_sources_sinks.pysa#L50-L55) wouldn't give us coverage for those subtypes, since Pysa wouldn't know about the relationship.

One thing that makes life a bit easier is that you don't need to fully specify the stubs. You can elide anything you don't care about with a `...`, and you can even just put an empty file in the `stubs` directory ([eg](https://github.com/facebook/pyre-check/blob/master/stubs/django/http/multipartparser.pyi)) to inform Pyre you don't care about type checking on the module with the same name as that file.

> After adding non-universal sanitizers, more sanitizers to the standard stubs/ directory would be helpful. (Already on the roadmap.)

 Yeah, adding that feature and fleshing out the sanitizers is definitely on my wishlist. There are already plenty of places where I'm adding features and filtering on them, where I could easily make them a non-universal sanitizer

PR #243, opened 2020-03-07
Unbreak Python CI

thatch:
It looks like the CI is failing right now, e.g. https://github.com/facebook/pyre-check/runs/491180735 and this should fix both the pyre and black steps.  I also made the selfcheck able to run against any version virtualenv (I tested on 3.8, and the CI will test on 3.6).

### DIFF
--- a/client/apply_annotations.py
+++ b/client/apply_annotations.py
@@ -89,8 +89,9 @@ class TypeCollector(cst.CSTVisitor):
 
     def visit_FunctionDef(self, node: cst.FunctionDef) -> bool:
         self.qualifier.append(node.name.value)
-        if node.returns is not None:
-            return_annotation = self._create_import_from_annotation(node.returns)
+        returns = node.returns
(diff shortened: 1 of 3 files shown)


gbleaney:
Thanks for the PR @thatch! We've been bad at keeping our github CI green (I swear we do have internal CI that passes though :) ), so I really appreciate this. I'll get this merged shortly

gbleaney:
Actually, it looks like the change to `profile.py` is failing our internal pyre run. I suspect this is because the released version of pyre is slightly behind what we run in our internal test. @thatch, do you mind removing your changes to that file, and we can instead try and fix that part of the breakage with a new release? CC: @MaggieMoss

thatch:
Reverted changes to `profile.py` now.

gbleaney:
Thanks again @thatch! This will be landing soon, and I'm working on doing the new release now

PR #260, opened 2020-05-10
Add more Django model fields

dirn:
This adds the following fields to `django.db.models.fields`

* `DecimalField`
* `FloatField`
* `SmallAutoField`
* `TimeField`
* `URLField`

PR #261, opened 2020-05-10
Add initial stubs for Django forms

dirn:
There's quite a bit more than can be annotated for `django.forms`, but
this is a jumping off point. It adds a handful of input types and a
couple of fields, along with `Form` and `ModelForm`.


gbleaney:
Sorry for the delay and thanks for the PR! Landing it now. I'm also going to be doing some re-syncing from our internal and open source repos, so you many notice the Django stubs shift slightly in the next day or to, but I don't think it'll conflict with #260 or further `forms` stubs.

PR #262, opened 2020-05-11
Add stubs for Django's gettext

dirn:
Django's `ugettext` family of translation functions are deprecated.
They've been aliases for the `gettext` equivalents since Django 2.0.


gbleaney:
Looks like this one was easier to land than the last. Passed all our internal tests on the first try :). Thanks for the PR!

dirn:
This one was pretty straightforward. If only they were all this easy.

PR #272, opened 2020-06-15
Add initial stubs for BaseUserManager and UserManager

dirn:
These classes should be used whenever the included `User` is extended or
replaced. Like the other stubs for this module, they just establish the
types.

Because `BaseUserManager` belongs in `base_user`, not `models`, the
module it being added and `AbstractBaseUser` is being moved there as
well.

### DIFF
--- a/stubs/django/contrib/auth/models.pyi
+++ b/stubs/django/contrib/auth/models.pyi
@@ -2,16 +2,17 @@
 
 from typing import Any, Optional
 
+from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
 from django.db import models
+
+class UserManager(BaseUserManager):
+    pass
 
 class AnonymousUser(object):
     id: Optional[int]
     is_active: bool
     def is_anonymous(self) -> bool: ...
     def is_authenticated(self) -> bool: ...
-
-class AbstractBaseUser(models.Model):
-    pass
 
 class AbstractUser(AbstractBaseUser):
     pass
--- a/stubs/django/contrib/auth/base_user.pyi
+++ b/stubs/django/contrib/auth/base_user.pyi
@@ -1,0 +1,9 @@
+# pyre-unsafe
+
+from django.db import models
+
+class BaseUserManager(models.Manager):
+    pass
+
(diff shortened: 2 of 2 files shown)


stubs/django/contrib/auth/models.pyi
```diff
@@ -19,4 +19,10 @@ class AbstractUser(AbstractBaseUser):
 class User(AbstractUser):
     pass
 
+class BaseUserManager(models.Manager):
```
gbleaney, line 22: It looks like this actually lives in `base_user.py`:
https://github.com/django/django/blob/e29637681be07606674cdccb47d1e53acb930f5b/django/contrib/auth/base_user.py#L16

Would you mind putting it there and importing it in this file instead?

dirn: It does, but so does [`AbstractBaseUser`](https://github.com/django/django/blob/e29637681be07606674cdccb47d1e53acb930f5b/django/contrib/auth/base_user.py#L47). That's why I put it here. Want me to move both?

gbleaney: Yeah, if you could that'd be great. These models have sometimes been evolved quickly to prevent typechecking errors, without looking at the original source. I guess in this case, everyone imported `AbstractBaseUser` from `django.contrib.auth.models`, so that's where we added the placeholder class.

gbleaney:
As a totally unrelated aside, mind if I ask what you're using these stubs for? I'm loving the PRs, and I'm curious what application you found for these. We had debated whether they were adding value for anyone or not

dirn:
I've been working on a [site for financial aid applications for PyGotham](https://github.com/PyGotham/awards). All of the PRs I've sent here either are things used in that project or are things directly related to things used there.

The code is littered with `pyre-ignore` comments that reference my PRs. If nothing else, they're adding value for me.

gbleaney:
Awesome, I'm excited that it's adding value for you :). I wonder if we should reignite the debate about trying to clean up these stubs and upstream them to typeshed.

Landing your PR now. As usual, thanks for taking the time to make these!

PR #273, opened 2020-06-15
Elsewhere Pysa is called the Pyre Static Analyzer

muglug:


gbleaney:
Hey @muglug thanks for spotting the inconsistency! We had a discussion internally and decided that we prefer calling it "Python Static Analyzer". I've made the change to the other places where we were calling it "Pyre Static Analyzer" internally, and you should see it showing up in the repo and docs soon.

I'm going to close this PR, since we chose to make the change elsewhere.

gbleaney:
Yeah, you're totally right about “static analysis” being redundant. Something like "Taint Flow Analysis" would probably be more accurate and less redundant.

"Python Security Analyzer" is an awesome bacronym to better capture what Pysa does currently. The reason that I hesitate to use that, however, is because Pysa is actually useful for more than just security. We don't have resources on those other applications in our external docs, but we've found other uses for it internally. It can really be helpful for any problem that can be modeled as data flows.

muglug:
Yeah, I get that, but its _main_ use case is for detecting security vulnerabilities and, as you say, it’s just an acronym anyway.

But now I’m curious - what sorts of (non-security-related) problems have you been able to model with dataflow analysis?

gbleaney:
We haven't talked publicly about other applications, so I can't go into detail, but here is one interesting example:

What if you detected a performance regression in a given function, but that code itself hadn't changed? One cause could be that some of the data used to make decisions in that function had changed. You could use data flow analysis to track data flows from experimentation frameworks or other external data sources into control flow operators in that function to automatically attribute the regression.

muglug:
fascinating, thanks!

PR #288, opened 2020-08-04
Add argument to __getattr__ stubs

### LINKED ISSUE #286: Pyre can't find names using __getattr__
When trying to import from a module that uses `__getattr__` in its stub file, `pyre` can't find the name. For example

    from django.contrib.auth.decorators import login_required

results in

    Undefined import [21]: Could not find a name `login_required` defined in module `django.contrib.auth.decorators`.

This import worked prior to 5f1d94d7d29207974145427086bd7bb97f379155. I'm running into this in PyGotham/awards#78, but I'm able to reproduce this in a file that just contains the import.

dirn:
When Python calls a module's `__getattr__` function, it passes it the
name of the attribute. Without including the argument in the signature,
Pyre can't find module attributes not explicitly included in the stubs.
This results in `Undefinied import` errors.

To address this, a `name` argument (typed as `str`) is being added to
each `__getattr__` stub.

Fixes #286


arthaud:
Hi @dirn, thanks for the contribution!

Actually, I have already done this in 0dd5c8f1622b4d7f49b196e26b018c6f7de2297d.
Sorry for the waste of time!

dirn:
No worries. I checked `master` and didn't see any PRs.

PR #289, opened 2020-08-04
Export redirected Django imports

### LINKED ISSUE #287: Pyre can't find redirected imports
When trying to import something is actually a redirected import, `pyre` can't find the name. For example

    from django.urls import reverse

results in

    Undefined import [21]: Could not find a name `reverse` defined in module `django.urls`.

whereas

    from django.urls.base import reverse

passes.

I'm running into this in PyGotham/awards#78, but I'm able to reproduce this in a file that just contains the import.

dirn:
As per [PEP 484], an import in a stub file isn't considered exported
from the stub unless the import includes `as`. This will address this
for all of the imports that I could easily tell were intended to be
exported.

Fixes #287

[PEP 484]: https://www.python.org/dev/peps/pep-0484/#stub-files

### DIFF
--- a/stubs/django/core/mail/__init__.pyi
+++ b/stubs/django/core/mail/__init__.pyi
@@ -3,7 +3,11 @@
 from email.mime.base import MIMEBase
 from typing import Any, Iterable, Optional, Tuple, Union
 
-from django.core.mail.message import EmailMessage, SafeMIMEMultipart, SafeMIMEText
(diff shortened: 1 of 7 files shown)


arthaud:
Thanks for the contribution!

I was already working on this and pushed 676acecf5d9308241796cd8b2eacb71a94d46b66.

PR #307, opened 2020-08-10
Updated pysa_precollectors to point to the correct exercise.

robertTheHub:
The exercise pointed to in this doc is currently for Features and SAPP instead of Model Generators.


gbleaney:
Awesome catch! We added a new exercise in the middle, and forgot to update this link

sathish798:
How may I contact ur

PR #312, opened 2020-09-03
Fix to docs CI (Actions)

brianjo:
Removes require.resolve option for fb/sidebars.js file. 
This should allow the build to complete. Tested on fork.


gbleaney:
I think is was already resolved another way, but thanks for the PR!

PR #320, opened 2020-10-12
Import Any inside placeholder Django stubs

dirn:
A lot of the placeholder stubs for Django use

    def __getattr__(name: str) -> Any: ...

to allow typechecks to pass before the module has actually been typed.
Unfortunately this can introduce errors stating that `Any` isn't defined.

This change will import `Any` into all of the placeholder stubs to avoid
this error.


gbleaney:
Oof, looks like a silly oversight on our part. Thanks for fixing that!

PR #321, opened 2020-10-12
Add empty stubs for Django __init__ files

dirn:
While Pyre seems to handle the absence of `__init__.pyi` well, there can
be problems with other type checkers that use its Django stubs. Adding
these stub files makes the others discoverable, much like adding empty
`__init__.py` files to Python packages.


dirn:
@gbleaney I rebase this on top of the latest version of `master`. Let me know if there are any changes that need to be made based on [your Phabricator diff][diff].

[diff]: https://www.internalfb.com/intern/diff/D24278299/

gbleaney:
Sorry for losing track of this @dirn! I've re-imported and will try to land today

PR #324, opened 2020-10-13
Make exercise 5 easier to uncomment.

xmo-odoo:
Exercise 5 requires uncommenting some lines, but it uses both line comments and *inline* comments which makes it somewhat fiddly: most editors have a command to (un)comment lines or entire blocks at once, not so for inline comments.

However the inline comments seem unnecessary: they just comment trailing commas, which Python is perfectly fine with. By removing those, it becomes possible to just point to or select the offending line and bulk-uncomment them via whatever editor feature does that, which is quite a bit simpler.


gbleaney:
Good suggestion! I honestly have no idea why I commented those out... Maybe my C+03 is showing?

PR #328, opened 2020-10-15
Upgrade black to match internal version

jreese:


dkgi:
I think we're now setting the current version in our GitHub actions.

PR #333, opened 2020-11-06
Add stubs for Django's database errors

dirn:
These are the errors that are exported through `django.db`.
`OperationalError` was already included in the package's stubs, so it's
being moved into the stubs for the `utils` module since it's defined
there like the others.


gbleaney:
Thanks as usual for the PR. Will get this merged soon!

PR #334, opened 2020-11-06
Update miscellaneous Django stubs

dirn:
This will add some new types, update some existing types, add some
placeholders.


stubs/django/test/__init__.pyi
```diff
@@ -11,6 +11,5 @@ from django.test.client import Client as Client
 from django.test.utils import override_settings as override_settings
 
 class SimpleTestCase(TestCase):
-    client: testing.client.base.TestClient = ...
```
dirn, line 14: To whomever reviews this, `testing` isn't imported by this stub. As best I can tell, `client` is set by `test_client`, which defaults to `django.test.client.Client` but can take any class that has the same API. Most likely this will be a subclass of `Client`, but if you think an interface is better here, let me know. (Perhaps that's what `TestClient` is intended to be.)

stubs/django/test/__init__.pyi
```diff
@@ -11,6 +11,5 @@ from django.test.client import Client as Client
...
-    @contextlib.contextmanager
```
gbleaney, line 15: Did you intend to delete this line? Looks like they do say `settings` is a context manager:
https://github.com/django/django/blob/dbb4a86fa7c15fb9cf73a5954356eb88d65ac25f/django/test/testcases.py#L307

gbleaney (commented):
One question. Otherwise LGTM

stubs/django/test/__init__.pyi
dirn: My attention was drawn here by the missing import. `settings` itself isn't defined as a context manager; its return value is. That's what tripped me up. I'll add the import and put the decorator back.

PR #337, opened 2020-11-11
Create configuration from virtual environment

dkgi:


dkgi:
Will do a second pull request on the pyre repository itself rather than my own fork...

PR #362, opened 2021-01-05
Also run usort in linting

dkgi:


dkgi:
Doesn't quite work as expected. Abandoning for now...

PR #379, opened 2021-02-22
updated Contributing docs

m0mosenpai:
Upon building project from source, `_build` is  dumped in `pyre-check/source` and not in the root directory itself. Hence, example export command for `PYRE_BINARY` could be updated.


gbleaney:
Thanks for the fix! Yet another relic of our folder re-organization. I'll get this landed soon.

PR #382, opened 2021-02-22
Only require `dataclasses` for python3.6

rcramblit:
`dataclasses` is included with the stdlib for python>=3.7, so it should only need to be installed for python3.6 environments.

`dataclasses==0.6` is technically compatible with python3.7, so it is being installed when installing this package, but that clashes with other projects (namely `Black`, which requires `dataclasses>=0.8` in a python3.6 environment, and uses the built-in `dataclasses` in python3.7+ environments). To avoid jumping through hoops specifying these dependencies down the chain with pip, this package should avoid installing `dataclasses` outside of a python3.6 environment.


gbleaney:
Thanks for the PR! I'll get this landed soon

PR #386, opened 2021-03-03
Updated installation.md

m0mosenpai:
MakeFile is dumped in the source directory now, instead of the checkout (root) directory.


gbleaney:
Thanks! I really appreciate you cleaning up some of the rot in our doc. I've kicked off the land process; the diff should auto close soon.

PR #389, opened 2021-03-07
Renamed Server for Pyre to PyreServer

saydus:
Renaming the language server for Pyre to PyreServer for less ambiguity before introducing PysaServer. Pyre and Pysa will have different language servers, because they're going to handle different files, and be used by different users.

### DIFF
--- a/client/commands/v2/persistent.py
+++ b/client/commands/v2/persistent.py
@@ -237,7 +237,7 @@ class ServerState:
     )
 
 
-class Server:
+class PyreServer:
     # I/O Channels
     input_channel: connection.TextReader
     output_channel: connection.TextWriter
@@ -773,7 +773,7 @@ async def run_persistent(
             client_capabilities = initialize_result.client_capabilities
             LOG.debug(f"Client capabilities: {client_capabilities}")
             initial_server_state = ServerState()
-            server = Server(
+            server = PyreServer(
                 input_channel=stdin,
                 output_channel=stdout,
                 client_capabilities=client_capabilities,
--- a/client/commands/v2/tests/persistent_test.py
+++ b/client/commands/v2/tests/persistent_test.py
@@ -29,7 +29,7 @@ from ..persistent import (
     InitializationSuccess,
     InitializationFailure,
     InitializationExit,
-    Server,
+    PyreServer,
     ServerState,
     parse_subscription_response,
     SubscriptionResponse,
@@ -212,7 +212,7 @@ class PersistentTest(testslide.TestCase):
     @setup.async_test
     async def test_open_close(self) -> None:
         server_state = ServerState()
-        server = Server(
+        server = PyreServer(
             input_channel=create_memory_text_reader(""),
             output_channel=create_memory_text_writer(),
             client_capabilities=lsp.ClientCapabilities(),
@@ -276,7 +276,7 @@ class PersistentTest(testslide.TestCase):
         )
         bytes_writer = MemoryBytesWriter()
         fake_task_manager = BackgroundTaskManager(WaitForeverBackgroundTask())
-        server = Server(
+        server = PyreServer(
             input_channel=create_memory_text_reader(""),
             output_channel=TextWriter(bytes_writer),
             client_capabilities=lsp.ClientCapabilities(),
@@ -314,7 +314,7 @@ class PersistentTest(testslide.TestCase):
     @setup.async_test
     async def test_open_triggers_pyre_restart(self) -> None:
         fake_task_manager = BackgroundTaskManager(WaitForeverBackgroundTask())
-        server = Server(
+        server = PyreServer(
             input_channel=create_memory_text_reader(""),
             output_channel=create_memory_text_writer(),
             client_capabilities=lsp.ClientCapabilities(),
@@ -340,7 +340,7 @@ class PersistentTest(testslide.TestCase):
     @setup.async_test
     async def test_open_triggers_pyre_restart__limit_reached(self) -> None:
         fake_task_manager = BackgroundTaskManager(WaitForeverBackgroundTask())
-        server = Server(
+        server = PyreServer(
             input_channel=create_memory_text_reader(""),
             output_channel=create_memory_text_writer(),
             client_capabilities=lsp.ClientCapabilities(),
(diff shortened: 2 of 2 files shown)


gbleaney:
Hey @saydus, can you include some evidence of testing here? I'd showing that you can still launch and query the persistent server via VSCode would be enough.

saydus:
I ran into issues with local run of the extension (even with the current `master` codebase). I will try debug and fix it today, before our meeting tomorrow.

saydus:
Just successfully ran from source to test it out!

Here is how output looks like in the problems tab when I run the extension:
Screen Shot 2021-03-22 at 12 43 33 AM

gbleaney:
Thanks for adding that testing. I've imported this to go through our internal review system

arthaud:
Could you please rebase this on top of master? Thanks.

saydus:
@arthaud, rebased my branch from current master. It's one commit ahead now. Let me know if this works!

gbleaney:
@grievejia can you confirm whether the testing on this one works, or whether it's also using the old commands? It seems like the current VSCode plugin in the repo only invokes `pyre persistent`, so I'm assuming that means it's using the old commands?

grievejia:
> @grievejia can you confirm whether the testing on this one works, or whether it's also using the old commands? It seems like the current VSCode plugin in the repo only invokes `pyre persistent`, so I'm assuming that means it's using the old commands?

Yeah it seems so. Also according to the screenshot, Pyre was not able to properly find typeshed and other related search paths -- it's less clear to me what happened there. 

To test out the plugin, I would not recommend using the Pyre project itself. Instead, I suggest the following:
- Create a fresh new project directory somewhere else and `cd` to it.
- Create some test `.py` files under the directory (preferably introducing some type errors intentionally) and run `pyre init` there to set up an initial `.pyre_configuration`. I suspect that you need to say "yes" when being asked about initializing watchman for now, so `pyre persistent` could work properly later. 
- Run `pyre check` and verify that the type errors you see are reasonable.
- Add a `"use_command_v2": true` option to the generated `.pyre_configuration` file. 
- Run `pyre persistent 2>log.txt` on the command line and see if the LSP server would start and function properly (check `log.txt` for logging). 
- Once you have verified that `pyre persistent` works on the cli, fire up VSCode and open those test `.py` files there. Verify that type errors are showing up.

saydus:
@grievejia, Thank you so much for a detailed response! 

I first wanted to test it out on current `master`, so here is the result (sorry it's very long, but I tried to include all details):

### Test Code
 First off, here is the test code I used to type check:
```python
def greet(name: str) -> str:
    return "Hello " + name

print(greet(2))
```
### Pyre Configuration for test code

Here is my `.pyre_configuration`
```json
{
  "binary": "/Users/saydus/Desktop/test/pyre-check/source/_build/default/main.exe",
  "source_directories": [
    "."
  ],
  "typeshed": "/Users/saydus/Desktop/test/pyre-check/stubs/typeshed",
  "use_command_v2": true,
  "workers": 8
}
```
I set the workers to 8 to avoid getting messages that would indicate that number of workers was inferred to be 8 (it would print a lot of it in logs).

### Running `pyre check`
I am not sure if the result is expected, but this is the messages that I get from the above test code:
```bash
➜  pyre-testing pyre check
ƛ Using virtual environment site-packages in search path...
ƛ Found 2 type errors!
main.py:1:19 Undefined or invalid type [11]: Annotation `str` is not defined as a type.
main.py:2:11 Undefined attribute [16]: `str` has no attribute `__add__`.
``` 
Screen Shot 2021-03-26 at 1 00 45 AM

Pyre doesn't seem to recognize a built-in type. I tested it both with and without `"use_command_v2": true` and got same output. 

### Running `pyre persistent 2>log.txt`
I ran this command both with and without `"use_command_v2": true` in configuration. The `log.txt` output is as follows:

**With `"use_command_v2": true`:**
```

ƛ Using virtual environment site-packages in search path...


ƛ 
```
Hm, I am not sure if the log should be empty, but at least it doesn't fail. Should I expect any output in logs? I ran it without `"use_command_v2": true` and the logs were richer, even though it was constantly trying to connect to the monitor.

I also noticed that running it created a `.pyre` directory which I delete right after running for it to be "clean" for further tests.


### Testing from the extension
I opened the test directory in VSCode in debug mode for current `pysa-vscode`, which mirrors the behavior of `vscode` for Pyre but was initialized using newer boilerplate generator for VSCode extensions (uses newer TypeScript, etc.). 

It first failed, but going through logs in `.pyre`, I realized I needed to `pyre stop` in terminal after my testing.

Screen Shot 2021-03-26 at 1 49 11 AM

It is consistent with what I had in the terminal!

## Testing PR's branch now
Now, after going through this for the current `master` branch, I will repeat the process for `saydus:refactor-pyre-language-server` branch from this PR.

1. Running from `pyre check`
Screen Shot 2021-03-26 at 1 11 24 AM

2. Running `pyre persistent 2>log.txt`
```
ƛ Using virtual environment site-packages in search path...


ƛ 
```
Same output!

3. Running the extension
Screen Shot 2021-03-26 at 1 55 25 AM
Consistent with what I had for `master` and with terminal output!


## Conclusion
So behavior for both current master and PR's branch is identical. However, I am not sure if Pyre's output is correct -- let me know if this is what's expected!

saydus:
Oh, I just realized that `typeshed` was a Zip file in the directory. I unzipped it, and provided the correct path to typeshed, and here is the result:

Screen Shot 2021-03-26 at 2 07 40 AM

Screen Shot 2021-03-26 at 2 08 31 AM


Finally, it all worked as expected!

gbleaney:
@saydus Thanks for the detailed testing. That looks awesome to me. @grievejia any objection to me shipping?

grievejia:
That was some really good work @saydus! Let's ship this 🚢 !

> I am not sure if the log should be empty, but at least it doesn't fail. Should I expect any output in logs?

Sorry looks like `pyre persistent` did not set up the logging format properly. I'll try to fix that separately. Thanks for letting me know! As long as it's not crashing, things should be fine for now. 

On a side note: what does `pip show pyre-check` says on your machine? Just want to confirm if you are on the latest pyre version of `0.9.0`.

PR #393, opened 2021-03-16
Creating Pysa language server (temporary copy of Pyre's server)

saydus:
Creating a copy of Pyre's language server and server handler for Pysa. After having two separate language server classes, I will introduce logic for the `-tool` command that will use a different language server based on arguments passed with the `persistent` command.


gbleaney:
Hey @saydus, same comment as #389: Can you add some evidence of testing here?

saydus:
To test this one out, I just swapped Server with PysaServer and passed PysaServerHandler instead of PyreServerHandler [here](https://github.com/facebook/pyre-check/blob/996e4552421fbe5dacb707dc5cb4947d1dbc9283/client/commands/v2/persistent.py#L776). I ran `pyre persistent` before and after the change to get same results.

Screen Shot 2021-03-22 at 1 18 44 AM

This is not a breaking change because `PysaServer` is not used anywhere. However, I will introduce logic for branching into using separate servers in next PR for `-tool` flag.

gbleaney:
Thanks for the testing. Hopefully we can land soon!

client/commands/v2/persistent.py
```diff
@@ -397,6 +397,168 @@ async def run(self) -> int:
...
+                    return 0
+                else:
+                    raise json_rpc.InvalidRequestError("LSP server has been shut down")
+
+    async def _try_restart_pyre_server(self) -> None:
+        if self.state.consecutive_start_failure < CONSECUTIVE_START_ATTEMPT_THRESHOLD:
+            await self.pyre_manager.ensure_task_running()
+        else:
+            LOG.info(
+                "Not restarting Pysa since failed consecutive start attempt limit"
```
gbleaney, line 443: FYI, we just added the FixIt linter internally, which I don't think is running on the open source CI. It doesn't like implicit string concatenation. This will need to be fixed here and in a few other places (I'll comment for now, but you should try to get FixIt up and running in your environment. It's this rule specifically that's biting you: https://fixit.readthedocs.io/en/latest/rules/UsePlusForStringConcatRule.html
```suggestion
                ("Not restarting Pysa since failed consecutive start attempt limit"
                + " has been reached.")
```

client/commands/v2/persistent.py
```diff
@@ -743,6 +905,245 @@ async def run(self) -> None:
...
+            LOG.warning(message)
+        elif level == lsp.MessageType.INFO:
+            LOG.info(message)
+        else:
+            LOG.debug(message)
+        await self.show_message_to_client(message, level)
+
+    def update_type_errors(self, type_errors: Sequence[error.Error]) -> None:
+        LOG.info(
+            "Refereshing type errors received from Pysa server. "
```
gbleaney, line 957: Here too

client/commands/v2/persistent.py
```diff
@@ -743,6 +905,245 @@ async def run(self) -> None:
...
+            f"Total number of type errors is {len(type_errors)}."
+        )
+        self.server_state.diagnostics = type_errors_to_diagnostics(type_errors)
+
+    async def show_type_errors_to_client(self) -> None:
+        for path in self.server_state.opened_documents:
+            await _publish_diagnostics(self.client_output_channel, path, [])
```
gbleaney, line 964: https://fixit.readthedocs.io/en/latest/rules/GatherSequentialAwaitRule.html

client/commands/v2/persistent.py
```diff
@@ -743,6 +905,245 @@ async def run(self) -> None:
...
+        socket_path = server_connection.get_default_socket_path(
+            log_directory=Path(self.pyre_arguments.log_path)
+        )
+        try:
+            async with connection.connect_in_text_mode(socket_path) as (
+                input_channel,
+                output_channel,
+            ):
+                await self.log_and_show_message_to_client(
+                    "Established connection with existing Pyre server at "
```
gbleaney, line 1046: Here too

client/commands/v2/persistent.py
```diff
@@ -743,6 +905,245 @@ async def run(self) -> None:
...
+                    event=LSPEvent.CONNECTED,
+                    normals={
+                        "connected_to": "already_running_server",
+                        **self._auxiliary_logging_info(),
+                    },
+                )
+                await self.subscribe_to_type_error(input_channel, output_channel)
+        except connection.ConnectionFailure:
+            await self.log_and_show_message_to_client(
+                f"Starting a new Pysa server at `{self.server_identifier}` in "
```
gbleaney, line 1061: Here too

client/commands/v2/persistent.py
```diff
@@ -743,6 +905,245 @@ async def run(self) -> None:
...
+                    _log_lsp_event(
+                        remote_logging=self.pyre_arguments.remote_logging,
+                        event=LSPEvent.NOT_CONNECTED,
+                        normals={
+                            **self._auxiliary_logging_info(),
+                            "exception": str(start_status.detail),
+                        },
+                    )
+                    await self.show_message_to_client(
+                        f"Cannot start a new Pyre server at `{self.server_identifier}`.",
```
gbleaney, line 1102: https://github.com/PyCQA/flake8-bugbear#list-of-warnings B950

client/commands/v2/persistent.py
```diff
@@ -743,6 +905,245 @@ async def run(self) -> None:
...
+                        # The heuristic here is that if the restart has failed
+                        # this many times, it is highly likely that the restart was
+                        # blocked by a defunct socket file instead of a concurrent
+                        # server start, in which case removing the file could unblock
+                        # the server.
+                        LOG.warning(f"Removing defunct socket file {socket_path}...")
+                        stop.remove_socket_if_exists(socket_path)
+                else:
+                    await self.show_message_to_client(
+                        f"Pyre server restart at `{self.server_identifier}` has been "
```
gbleaney, line 1119: Here too

gbleaney (changes requested):
Looks like you're getting bit by some linter errors. Setting up flake8 and fixit will probably help catching these comments

grievejia:
@saydus Thanks for the patch! A couple of issues here:
- Would you mind moving the pysa server into a separated file? The `persistent.py` is already lengthy. 
- Unfortunately your test at [this comment](https://github.com/facebook/pyre-check/pull/393#issuecomment-803794866) does not actually test what you expected. From the screenshot, it is highly likely that you were invoking the legacy persistent command, which is orthogonal to any of the logic added in this commit. To test the `v2/persistent.py` logic, you need to run `pyre --use-command-v2 persistent` or add `"use_command_v2": true` in the pyre configuration of your test project. 
- I am actually quite surprised that you choose to copy and paste everything from the Pyre LSP server to the Pysa server, since a large part of the Pyre LSP server implementation is Pyre-specific. For example, why does the Pysa server include the same `ServerState` as the Pyre server? I presume they are going to track very different states? Also, the `PyreServerHandler` class is implemented under the assumption that Pyre supports both a server mode and an error subscription-based workflow, neither of which is true for Pysa. So I am quite surprised that you copied the code into this `PysaServerHandler` class as well. Could you explain a bit of your thoughts on these decisions?

saydus:
Thank you so much for clarifying the confusion with commands @grievejia. I will test it out separately and move the Pysa's server into a separate file. 

Sorry for a possible miscommunication for why the Pysa LSP just mirrors Pyre - I was following the steps outlined [here](https://github.com/MLH-Fellowship/pyre-check/issues/3) by @gbleaney. I think the decision to temporarily have the language server as an exact copy was made to remove the blocker for adding the logic for `tool` flag for `persistent` command. As the flag gets added and we can connect to the the server from two different VSCode plugins, the next step will be implementing actual Pysa's LSP.

saydus:
I modified implicit string concatenation that was present in the code. I want to note that this was just a copy of existing Server for Pyre, which also produces these linting issues.

For introducing the Server in a separate file, I had questions about what to do with some of the repeating dependencies. For example `LSPEvent` enum [here](https://github.com/facebook/pyre-check/blob/8bc3eab39306a7f0f11839ea6958a359a2616c68/client/commands/v2/persistent.py#L45) would be shared by two servers. I think it would make sense to separate both servers into a separate file, where they would share such dependencies. If so, I was wondering if this should be a separate refactor PR instead of doing it here. 

For linting, what do you suggest I use not to get these errors? Trying to check for lint errors while I am editing the files produces a large amount of errors because it seems like older code didn't use FixIt.

grievejia:
> For introducing the Server in a separate file, I had questions about what to do with some of the repeating dependencies. For example LSPEvent enum here would be shared by two servers. I think it would make sense to separate both servers into a separate file, where they would share such dependencies. If so, I was wondering if this should be a separate refactor PR instead of doing it here.

If in the Pysa server file wants to reuse part of the code inside `persistent.py`, what you can do is to `import` the class/function into the new file. I don't have strong opinions on whether to do it immediately or just copy&paste code right now and do the import later. The bottom line here is that code reuse is not necessarily affected by the decision of whether to create a new file or not. 

> For linting, what do you suggest I use not to get these errors? Trying to check for lint errors while I am editing the files produces a large amount of errors because it seems like older code didn't use FixIt.

I see. Thanks for sharing the experience! Given the situation, let's not worry too much about lint errors then -- I'll keep that in mind when reviewing patches in the future.

saydus:
@grievejia thanks for clarification! I guess my concern was that we would have to import `PysaServer` into `persistent.py` and I was not sure if `PysaServer` can depend on imports from `persistent.py` itself because it would lead to circular dependency.

grievejia:
> @grievejia thanks for clarification! I guess my concern was that we would have to import `PysaServer` into `persistent.py` and I was not sure if `PysaServer` can depend on imports from `persistent.py` itself because it would lead to circular dependency.

Sorry I don't quite get this -- why would you import `PysaServer` back into `persistent.py`? I believe the direction I was pushing for is to avoid that, and have this new file provide an entry point independent of what's in`persistent.py`.

saydus:
I went ahead and moved `PysaServer` to a separate file. In `persistent.py`, I imported Pysa's temporary server with its handler:
```python
from . pysa_server import (
    PysaServer,
    PysaServerHandler
)
```
My `pysa_server.py` file depends on some classes within `persistent.py`, so I tried importing them here as:
```python
from . persistent import (
    LSPEvent,
    CONSECUTIVE_START_ATTEMPT_THRESHOLD,
    ServerState
)
```

This is causing a circular dependency issue as expected. To temporarily test out if copy-paste would at least work, I redefined these classes in `pysa_server.py`. After redefining and changing `persistent.py` to use `PysaServer` with `PysaServerHandler`, I tested out our v2 `persistent` command, and it worked just fine! The logs are empty but the connection worked both in terminal and the extension.

Since this server's implementation is temporary anyway, I am not sure if I should just redefine these dependencies (they will be cleared anyway). I will commit it with the working change, where dependencies are redefined because they will change and not going to be same anyway.

Do you want me to move PyreServer to a separate file too?

saydus:
Omg, I think I rebased from upstream/master, and for some reason, all the commits since the PR started are now appearing in the history of this PR. Do you know what I did wrong here?

saydus:
Found a work-around by changing the base to something else and back to master. It's showing the correct change now 😅.

saydus:
@grievejia sorry I just saw your message. I think I misunderstood you for moving the server to a separate file then. I was thinking `PysaServer` would be used somewhere in this function, where the branching will happen depending on the command flag. 

If that's not the case, I can go ahead and refactor. Where would we include logic for using Pysa's server instead of Pyre's server if not in `persistent.py`?

grievejia:
> @grievejia sorry I just saw your message. I think I misunderstood you for moving the server to a separate file then. I was thinking `PysaServer` would be used somewhere in this function, where the branching will happen depending on the command flag.
> 
> If that's not the case, I can go ahead and refactor. Where would we include logic for using Pysa's server instead of Pyre's server if not in `persistent.py`?

Sorry for not being clear earlier. What I have in mind is that we keep `persistent.py` intact, and have your Pysa server in a separate `.py` file and expose a `run` function as an entry point just like `persistent.run` (it can be a copy-and-paste of what's in `persistent.py` to begin with just like what you did for PysaServer). Then you can pick-and-choose between Pyre and Pysa entry points in the top-level `pyre.py` file. 

I would also suggest not reusing the pre-existing `pyre persistent` command with a flag, but to have a new command for the Pysa server instead.

saydus:
That makes sense! Thank you for clarifying. I will refactor this file to reflect this approach then. Have a great weekend!

saydus:
I have duplicated `run` and `run_persitent` in `pysa_server.py` now! To test it out, I have modified `pyre.py` to use `v2.pysa_server.run` instead of `v2.persistent.run`. 
Screen Shot 2021-03-27 at 10 46 04 PM

Then, I went through the testing process for the v2 of persistent command, as in the previous PR. The command works as expected! There are no logs except for the following: (as expected)

```

ƛ Using virtual environment site-packages in search path...


ƛ 
```

Please, let me know if I there are any changes I should make!

client/commands/v2/__init__.py
```diff
@@ -13,3 +13,4 @@
 from . import start  # noqa F401
 from . import stop  # noqa F401
 from . import validate_models  # noqa F401
+from . import pysa_server  # noqa F40
```
gbleaney, line 16:
```suggestion
from . import pysa_server  # noqa F401
```

client/commands/v2/pysa_server.py
```diff
@@ -0,0 +1,546 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+"""
+This is a temporary implementation of Pysa's langauge server. This is an exact copy
```
gbleaney, line 8:
```suggestion
This is a temporary implementation of Pysa's language server. This is an exact copy
```

client/commands/v2/__init__.py
```diff
@@ -13,3 +13,4 @@
 from . import start  # noqa F401
 from . import stop  # noqa F401
 from . import validate_models  # noqa F401
+from . import pysa_server  # noqa F401
```
gbleaney, line 16: This import should be sorted alphabetically

client/commands/v2/pysa_server.py
```diff
@@ -0,0 +1,546 @@
...
+of Pyre's langauge server, which lives in persistent.py.
```
gbleaney, line 9:
```suggestion
of Pyre's language server, which lives in persistent.py.
```

client/commands/v2/pysa_server.py
```diff
@@ -0,0 +1,546 @@
...
+# LICENSE file in the root directory of this source tree.
+
+
+"""
+This is a temporary implementation of Pysa's langauge server. This is an exact copy
+of Pyre's langauge server, which lives in persistent.py. 
+"""
+
+import asyncio
+import dataclasses
```
gbleaney, line 13: Tons of unused imports here:

- dataclass
- enum
- json
- subprocess
- tempfile
- Set
- Union
- List
- version x2
- statistics
- server_event

You can set up `flake8` to catch these

gbleaney (changes requested):

saydus:
Thanks for reviewing it! Made all the changes and sorry for the import style issues - I assumed black catches stuff like this.

PR #397, opened 2021-03-23
added alternate method to launch pyre when built from source

m0mosenpai:
Added new instructions - using a bash script to invoke Pyre instead of an alias. These instructions work well when working with the VSCode extension.


documentation/website/docs/installation.md
```diff
@@ -70,6 +71,21 @@ $ source ~/.bashrc
 ```
 You should be able to open a new shell and run `pyre -h` now, confirming `pyre` was set-up correctly. Any changes made to the Pyre Python client code should be immediately observable the next time you invoke `pyre`
 
+#### Testing changes for Plugin Development
+VSCode often has problems working with aliases. As a result, Pyre fails to start in the VSCode Extension Development Host during the debugging process. To get around this, instead of creating an alias, we can create an executable script called `pyre` in the `scripts` directory with the following lines
```
gbleaney, line 75:
```suggestion
VSCode will not pick up your shell aliases, so the alias step in the previous section will not work if you're doing VSCode Plugin development. To get around this, instead of creating an alias, we can create an executable script called `pyre` and place it in a directory in our `PATH`:
```

documentation/website/docs/installation.md
```diff
@@ -70,6 +71,21 @@ $ source ~/.bashrc
...
+
+```bash
+#!/bin/bash
+python -m client.pyre "$@"
+```
+Add the `pyre-check` directory to `PYTHONPATH`, the `pyre-check/scripts` directory to `PATH` and then use the command `pyre` to launch the client like before
+
+```bash
+$ echo "export PYTHONPATH=/path/to/pyre-check:$PYTHONPATH" >> ~/.bashrc
```
gbleaney, line 84: This can be moved into the script file, rather than modifying the user's `PYTHONPATH` globally

documentation/website/docs/installation.md
```diff
@@ -70,6 +71,21 @@ $ source ~/.bashrc
...
+$ echo "export PATH=/path/to/pyre-check/scripts:$PATH" >> ~/.bashrc
```
gbleaney, line 85: You should add quotes like we did for the alias example above, so that spaces in `PATH` doesn't break things

gbleaney (changes requested):

documentation/website/docs/installation.md
```diff
@@ -70,6 +71,21 @@ $ source ~/.bashrc
 ```
 You should be able to open a new shell and run `pyre -h` now, confirming `pyre` was set-up correctly. Any changes made to the Pyre Python client code should be immediately observable the next time you invoke `pyre`
 
+#### Testing changes for Plugin Development
+VSCode will not pick up your shell aliases, so the alias step in the previous section will not work if you're doing VSCode Plugin development. To get around this, instead of creating an alias, we can create an executable script called `pyre` and place it in a directory in our `PATH`:
+
+```bash
+#!/bin/bash
+export PYTHONPATH="/path/to/pyre-check:$PYTHONPATH"
+python -m client.pyre "$@"
```
gbleaney, line 80: Nit: This can be a one liner
```suggestion
PYTHONPATH="/path/to/pyre-check:$PYTHONPATH" python -m client.pyre "$@"
```

documentation/website/docs/installation.md
```diff
@@ -70,6 +71,21 @@ $ source ~/.bashrc
...
+```
+Add the `pyre-check` directory to `PYTHONPATH`, the `pyre-check/scripts` directory to `PATH` and then use the command `pyre` to launch the client like before
```
gbleaney, line 82: First part is no longer relavent.

```suggestion
Add the `pyre-check/scripts` directory to `PATH` (assuming you placed the above script in that directory) and then use the command `pyre` to launch the client like before
```

gbleaney (changes requested):
Also, can you confirm if you re-tested the workflow with your changes?

documentation/website/docs/installation.md
```diff
@@ -71,6 +72,21 @@ $ source ~/.bashrc
 ```
 You should be able to open a new shell and run `pyre -h` now, confirming `pyre` was set-up correctly. Any changes made to the Pyre Python client code should be immediately observable the next time you invoke `pyre`
 
+#### Testing changes for Plugin Development
+VSCode will not pick up your shell aliases, so the alias step in the previous section will not work if you're doing VSCode Plugin development. To get around this, instead of creating an alias, we can create an executable script called `pyre` and place it in a directory in our `PATH`:
+
+```bash
+#!/bin/bash
+export PYTHONPATH="/path/to/pyre-check:$PYTHONPATH"
```
gbleaney, line 80: Looks like you forgot to delete this line

gbleaney (changes requested):

documentation/website/docs/installation.md
```diff
@@ -71,6 +72,21 @@ $ source ~/.bashrc
...
+PYTHONPATH="/path/to/pyre-check:$PYTHONPATH" python -m client.pyre "$@"
+```
+Add the `pyre-check/scripts` directory to `PATH` (assuming you placed the above script in that directory) and then use the command `pyre` to launch the client like before
+
+```bash
+$ echo 'export PATH="/path/to/pyre-check/scripts:$PATH"' >> ~/.bashrc
```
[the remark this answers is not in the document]
m0mosenpai: I don't think we need to escape it ..? I did test it out ... I ran the command as it is and I get this in my `.bashrc`
image

which I think is correct

documentation/website/docs/installation.md
m0mosenpai: Yeah sorry about that

documentation/website/docs/installation.md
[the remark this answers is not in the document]
gbleaney: Ahh, string interpolation is only in double quotes. Perfect, thanks for adding evidence!

gbleaney (approved):

documentation/website/docs/installation.md
```diff
@@ -71,6 +72,20 @@ $ source ~/.bashrc
 ```
 You should be able to open a new shell and run `pyre -h` now, confirming `pyre` was set-up correctly. Any changes made to the Pyre Python client code should be immediately observable the next time you invoke `pyre`
 
+#### Testing changes for Plugin Development
+VSCode will not pick up your shell aliases, so the alias step in the previous section will not work if you're doing VSCode Plugin development. To get around this, instead of creating an alias, we can create an executable script called `pyre` and place it in a directory in our `PATH`:
+
+```bash
+#!/bin/bash
+export PYTHONPATH="/path/to/pyre-check:$PYTHONPATH" python -m client.pyre "$@"
```
arthaud, line 80: We don't need the `export` here.
`A=B cmd` runs `cmd` with the environment variable `A` set to `B`, regardless of whether `A` was exported on not.

documentation/website/docs/installation.md
```diff
@@ -71,6 +72,20 @@ $ source ~/.bashrc
...
+```
+Add the `pyre-check/scripts` directory to `PATH` (assuming you placed the above script in that directory) and then use the command `pyre` to launch the client like before
+
+```bash
+$ echo 'export PATH="/path/to/pyre-check/scripts:$PATH"' >> ~/.bashrc
```
arthaud, line 85: nitpick here: we don't need the `export` here either. `PATH` is always exported, otherwise, well, nothing would work.

arthaud (changes requested):

documentation/website/docs/installation.md
m0mosenpai: Oh I didn't know that! I'll make the change thanks

PR #400, opened 2021-03-27
Delete duplicate import of `version`

saydus:
Noticed a small issue with `version` being imported twice here. I deleted the second import and left the one on line 21.


arthaud:
Thanks for the contribution :)

PR #405, opened 2021-04-04
Added a new command for starting a pysa language server

saydus:
This change would create a `pysa-language-server` to connect to Pysa's language server. How should we handle versioning of commands here? Pysa server was introduced in v2 folder, so should we assume that the command works in v2? 

The `stop` command seems to be generic for all running servers (it iterates through socket files) -- should there be additional command for stopping Pysa's server?

### DIFF
--- a/client/pyre.py
+++ b/client/pyre.py
@@ -886,45 +886,33 @@ def pysa_language_server(context: click.Context, no_watchman: bool) -> int:
     configuration = configuration_module.create_configuration(
         command_argument, Path(".")
     )
-    if configuration.use_command_v2:
-        log.start_logging_to_directory(
-            # Always log to file regardless of whether `-n` is given
-            noninteractive=False,
-            log_directory=configuration.log_directory,
-        )
-        return v2.pysa_server.run(
-            configuration,
-            command_arguments.StartArguments(
-                changed_files_path=command_argument.changed_files_path,
-                debug=command_argument.debug,
-                enable_memory_profiling=command_argument.enable_memory_profiling,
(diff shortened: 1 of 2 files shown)


client/pyre.py
```diff
@@ -873,6 +873,60 @@ def persistent(context: click.Context, no_watchman: bool) -> int:
...
+                saved_state_project=command_argument.saved_state_project,
+                sequential=command_argument.sequential,
+                show_error_traces=command_argument.show_error_traces,
+                store_type_check_resolution=False,
+                terminal=False,
+                wait_on_initialization=True,
+            ),
+        )
+    else:
+        return run_pyre_command(
```
gbleaney, line 918: This looks like it's still calling the old pyre persistent command. Should this be here?

gbleaney (changes requested):

client/pyre.py
saydus: I am confused as to whether or not anything should execute in case of "v1" command call. Should we only allow v2? If so, I will just delete that else statement.

grievejia: > I am confused as to whether or not anything should execute in case of "v1" command call.  Should we only allow v2?

Yes. V1 is considered legacy code these days. We are still in the process of migrating away from it. New code can use V2 directly.

grievejia:
> The stop command seems to be generic for all running servers (it iterates through socket files) -- should there be additional command for stopping Pysa's server?

The `stop` command has nothing to do with LSP servers. LSP servers do not need to worry about stopping -- VSCode will kill them when it quits.

PR #406, opened 2021-04-06
Added Docker image for building from source

gracewgao:
Here's the Dockerfile and the associated documentation for building pyre-check from source within a Docker container. This setup has already been tested on my Apple M1, but let me know if there's anything else I should change/add!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,9 +82,10 @@ During analysis, each function will be processed into a control flow graph to re
 
 ### Getting Started
 - Install `opam` and run `opam switch install "4.10.2"`
-- run ` ./scripts/setup.sh`
+- Run ` ./scripts/setup.sh`
 - After the setup script runs successfully, you should be able to `cd` into `source` and run `make` to build the Pyre binary
 
+Alternatively, you can also set up using [the Docker image](https://pyre-check.org/docs/installation/#building-from-docker) to build from source.
 
 ### How do I test my OCaml changes against real code?
 When you run pyre on the command line, what runs under the hood is a shim which finds and runs a suitable pyre binary. This works well for production use, but isn't great for testing out your own build of Pyre.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+FROM ocaml/opam:debian-10-ocaml-4.11
+
+LABEL maintainer="gracewgao"
+
+USER root
+
+# installs system dependencies
+RUN apt update && apt upgrade -y \
+    && apt install python3 python3-pip pkg-config libsqlite3-dev build-essential -y \
+    && opam switch create 4.10.2
+
+# creates symlink for Python 3
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+# updates PATH variable for opam
+ENV PATH="/home/opam/.opam/4.10.2/bin:$PATH"
+
+# adds then installs Python dependencies
+ADD --chown=1000:1000 requirements.txt /home/opam/pyre-check/
+WORKDIR /home/opam/pyre-check
+RUN pip3 install -r requirements.txt
+
+# adds then unzips typeshed
+ADD --chown=1000:1000 stubs/typeshed /home/opam/pyre-check/stubs/typeshed
+WORKDIR /home/opam/pyre-check/stubs/typeshed
+RUN unzip typeshed.zip && chown -R 1000:1000 typeshed-master
+
+# adds then builds pyre-check
+ADD --chown=1000:1000 . /home/opam/pyre-check
+WORKDIR /home/opam/pyre-check
+RUN ./scripts/setup.sh --local
+WORKDIR /home/opam/pyre-check/source
+RUN make && make test
+
+# workaround to enable running local changes to pyre commands
+RUN echo '#!/bin/bash\n \
+    python -m client.pyre "$@" \n' >> /home/opam/pyre-check/scripts/pyre \
+    && chmod +x /home/opam/pyre-check/scripts/pyre
+
+# runs python tests
+WORKDIR /home/opam/pyre-check
+RUN ./scripts/run-python-tests.sh
+
+# updates environment variables to use pyre-check
+ENV PYRE_BINARY="/home/opam/pyre-check/source/_build/default/main.exe"
+ENV PYRE_TYPESHED="/home/opam/pyre-check/stubs/typeshed/typeshed-master"
+ENV PYTHONPATH="/home/opam/pyre-check:$PYTHONPATH"
+ENV PATH="/home/opam/pyre-check/scripts:$PATH"
+
+# switches to home directory and non-root user on launch
+USER 1000
+WORKDIR /home/opam/
+
+CMD /bin/bash
--- a/documentation/website/docs/installation.md
+++ b/documentation/website/docs/installation.md
@@ -86,6 +86,43 @@ $ echo 'PATH="/path/to/pyre-check/scripts:$PATH"' >> ~/.bashrc
 $ source ~/.bashrc
 ```
 
+## Building from Docker
+
+If you're having issues setting up or your OS is not yet supported, you can also use a Docker image. It runs [Debian GNU/Linux 10 (buster)](https://www.debian.org/) and builds pyre-check from source.
+
+Before starting, ensure that [Docker](https://docs.docker.com/get-docker/) is installed on your computer.
+
+1. Clone the pyre-check repository and navigate to the root directory.
+   ```bash
+   git clone https://github.com/facebook/pyre-check.git
+   cd pyre-check
+   ```
+
+2. Build the Docker image with the tag `pyre-check` (or another tag if you wish)
+   ```bash
+   docker build -t pyre-docker .
+   ```
+
+3. Run the new image in a new container `pyre-container` (or another name if you wish)
+   ```bash
+   docker run \
+   --name pyre-container \
+   -v /path/to/your/directory:/src \
+   -t -i \
+   pyre-check
+   ```
+   *Note: Launching the container will build and run all tests.*
+
+4. Inside the container, run any Pyre command now with `pyre`!
+
+   *Note: When initializing Pyre with `pyre init`, you may have to enter the following paths for the binary and typeshed:*
+   ```bash
+   ƛ No `pyre.bin` found, enter the path manually: /home/opam/pyre-check/source/_build/default/main.exe
+   ƛ Unable to locate typeshed, please enter its root: /home/opam/pyre-check/stubs/typeshed/typeshed-master
+   ```
+
+**For contributors:** Inside the Docker container, the added pyre-check directory is only editable by the root user. To contribute to Pyre, make edits in your local filesystem and rebuild the Docker by running Step 2, then running a new Docker container in Steps 3-4.
+
 ## Windows Subsystem for Linux (WSL) Install
 
 On *x86_64* Windows `pyre` can run via *Linux* using [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux).


Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+LABEL maintainer="gracewgao"
+
+SHELL ["/bin/bash", "-c"]
+
+# updates environment variables
+ENV PYRE_BINARY='/home/opam/pyre-check/source/_build/default/main.exe'
+ENV PYRE_TYPESHED='/home/opam/pyre-check/stubs/typeshed/typeshed-master'
+ENV PYTHONPATH='/home/opam/pyre-check:$PYTHONPATH'
+ENV PATH='/home/opam/.opam/4.10.2/bin:/home/opam/pyre-check/scripts:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH'
```
gbleaney, line 11: Wow, are `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` all really missing from the `PATH` by default?

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN echo -e 'export PYRE_BINARY=/home/opam/pyre-check/source/_build/default/main.exe \n \
```
gbleaney, line 13: Why do we need to both set these using ENV and put them in `.bashrc`? Shouldn't it be one or the other? If for some reason there is a process which doesn't inherit the variables through `ENV` but does pick them up from `.bashrc`, couldn't we just eliminate the duplicate `ENV` commands by moving to `RUN source ~/.bashrc`?

Dockerfile
```diff
@@ -0,0 +1,51 @@
+FROM ocaml/opam:debian-10-ocaml-4.11
+
+LABEL maintainer="gracewgao"
+
+SHELL ["/bin/bash", "-c"]
+
+# updates environment variables
+ENV PYRE_BINARY='/home/opam/pyre-check/source/_build/default/main.exe'
```
gbleaney, line 8: Nit: Let's define these at the end when these executables actually exist

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN echo -e 'export PYRE_BINARY=/home/opam/pyre-check/source/_build/default/main.exe \n \
+    export PYRE_TYPESHED=/home/opam/pyre-check/stubs/typeshed/typeshed-master \n \
+    export PYTHONPATH=/home/opam/pyre-check:$PYTHONPATH \n \
+    export PATH=/home/opam/pyre-check/scripts:$PATH' >> ~/.bashrc
+
+# installs system dependencies
+RUN sudo apt update && sudo apt upgrade -y \
+    && sudo apt install python3 python3-pip pkg-config -y \
+    && sudo apt-get install libsqlite3-dev build-essential -y \
```
gbleaney, line 21: Why `apt` in some cases and `apt-get` in others?

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# installs system dependencies
+RUN sudo apt update && sudo apt upgrade -y \
+    && sudo apt install python3 python3-pip pkg-config -y \
+    && sudo apt-get install libsqlite3-dev build-essential -y \ 
+    && opam switch create 4.10.2
+
+# downloads and installs pyre-check
+RUN git clone https://github.com/facebook/pyre-check \
+    && cd pyre-check \
```
gbleaney, line 26: Prefer `WORKDIR` over `cd`:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#workdir

(Apply this throughout)

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    export PATH=/home/opam/pyre-check/scripts:$PATH' >> ~/.bashrc
+
+# installs system dependencies
+RUN sudo apt update && sudo apt upgrade -y \
+    && sudo apt install python3 python3-pip pkg-config -y \
+    && sudo apt-get install libsqlite3-dev build-essential -y \ 
+    && opam switch create 4.10.2
+
+# downloads and installs pyre-check
+RUN git clone https://github.com/facebook/pyre-check \
```
gbleaney, line 25: Given that the current user will have to have this repo checked out to get this dockerfile, we should use the working context rather than pulling in from the web (this has the additional benefit of working with local changes):
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#understand-build-context

You will probably want to add a `.dockerignore` as well, so avoid pulling in build artifacts and the like:
https://docs.docker.com/engine/reference/builder/#dockerignore-file

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+ENV PYTHONPATH='/home/opam/pyre-check:$PYTHONPATH'
+ENV PATH='/home/opam/.opam/4.10.2/bin:/home/opam/pyre-check/scripts:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH'
+
+RUN echo -e 'export PYRE_BINARY=/home/opam/pyre-check/source/_build/default/main.exe \n \
+    export PYRE_TYPESHED=/home/opam/pyre-check/stubs/typeshed/typeshed-master \n \
+    export PYTHONPATH=/home/opam/pyre-check:$PYTHONPATH \n \
+    export PATH=/home/opam/pyre-check/scripts:$PATH' >> ~/.bashrc
+
+# installs system dependencies
+RUN sudo apt update && sudo apt upgrade -y \
```
gbleaney, line 19: Commands like this that take a long time and don't change often should be as early in the file as possible to take advantage of build caching:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN git clone https://github.com/facebook/pyre-check \
+    && cd pyre-check \
+    && ./scripts/setup.sh --local
+    
+RUN cd pyre-check/source \
+    && make \
+    && make test
+
+# creates scripts for python alias and pyre-check
+RUN sudo touch /usr/bin/python \
```
gbleaney, line 34: I think these four lines be done by creating a symlink? Eg. my `python` command currently is a symlink:
```
(venv) gbleaney-mbp:python_security gbleaney$ which python
/Users/gbleaney/python_security/venv/bin/python
(venv) gbleaney-mbp:python_security gbleaney$ ll /Users/gbleaney/python_security/venv/bin/python
lrwxr-xr-x  1 gbleaney  1876110778  7 Mar 31 23:43 /Users/gbleaney/python_security/venv/bin/python -> python3
```

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    && sudo chmod a+w+x /usr/bin/python \
+    && sudo echo -e '#!/bin/bash\n \
+    python3 "$@" ' > /usr/bin/python \
+    && sudo echo -e '#!/bin/bash\n \
+    python -m client.pyre "$@" \n' >> /home/opam/pyre-check/scripts/pyre \
+    && sudo chmod +x /home/opam/pyre-check/scripts/pyre
+
+# runs python tests
+RUN cd /home/opam/pyre-check \ 
+    && pip3 install -r requirements.txt \
```
gbleaney, line 44: Same comment as on the `apt` step. Installing dependencies should probably go earlier

CONTRIBUTING.md
```diff
@@ -85,6 +85,39 @@ During analysis, each function will be processed into a control flow graph to re
 - run ` ./scripts/setup.sh`
 - After the setup script runs successfully, you should be able to `cd` into `source` and run `make` to build the Pyre binary
 
+### Docker Setup
+
+If you're having issues setting up or your OS is not yet supported, you can also use a Docker image. It runs [Debian GNU/Linux 10 (buster)](https://www.debian.org/) and builds pyre-check from source.
```
gbleaney, line 90: Let's not duplicate docs like this. Instead, let's put a pointer here to the other docs. That helps users find them, without adding extra maintenance overhead

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    && ./scripts/run-python-tests.sh
+
+# unzips typeshed
+RUN cd /home/opam/pyre-check/stubs/typeshed \
+    && unzip typeshed.zip
+
+CMD /bin/bash
```
gbleaney, line 51: Is this actually needed? I see in the instructions below, you have people also specify `/bin/bash` as the shell to use

documentation/website/docs/installation.md
```diff
@@ -86,6 +86,40 @@ $ echo 'PATH="/path/to/pyre-check/scripts:$PATH"' >> ~/.bashrc
...
+   docker run \                           
+   --name pyre-container \
+   -v /path/to/your/directory:/src \
+   -t -i \
+   pyre-check /bin/bash
+   ```
+
+4. Inside the container, run any pyre-check command now with `pyre`!
+
+   *Note: When initializing pyre-check with `pyre init`, you may have to enter the following paths for the binary and typeshed:*
```
gbleaney, line 117: Shouldn't these have been set in the dockerfile? Are these still required here?

gbleaney (changes requested):
Thanks for the PR! I'm really excited to have this landed as an alternative way to build pyre

Dockerfile
gracewgao: I'm not sure but it seemed unable to find programs like make etc. unless I did this? I think it's maybe because the $PATH variable isn't evaluating properly

Dockerfile
gracewgao: I tried that first but it wasn't working, so I found some way online to define the environment variables in Docker using ENV - I added them into .bashrc afterwards because I figured they would need to be reevaluated every time they restart the container?

Dockerfile
gracewgao: Yep, I'll change that!

Dockerfile
gracewgao: Honestly not too sure - which one would you recommend using?

Dockerfile
gracewgao: Thanks, I'll change that!

documentation/website/docs/installation.md
gracewgao: I think you have to set it up in the Dockerfile but then it asks for it again when initializing pyre?

Dockerfile
gbleaney: @gracewgao I think the problem is that you're using single quotes, not double quotes. You need double quotes for string interpolation to occur.

Dockerfile
gbleaney: It looks like variables set using `ENV` are supposed to persist when the container is run:
https://docs.docker.com/engine/reference/builder/#env

Dockerfile
gbleaney: IIRC `apt` is the newer, simpler, more user friendly wrapper. I don't have a strong preference either way, but I would prefer we be consistent

documentation/website/docs/installation.md
gbleaney: Ugh, we should fix that on the Pyre side, but that's out of scope here

Dockerfile
gracewgao: We can't install the dependencies until the requirements.txt inside the repo is cloned, so it would still have to be below that? I put them where they are now because the pip dependencies aren't really used for anything other than the Python tests, but I can change that if you want.

gbleaney: Generally you want expensive, infrequently changed steps higher in the file. So I'd say the correct order would be
1. apt installs
2. bring in requirements.txt
3. pip install
4. bring in typeshed and unzip
4. bring in rest of repo
5. build

Doing it that way means you never have to re-run the install or unzip steps, even if you're actively editing and changing code in the repo and re-launching the container

gracewgao: Okay, that makes sense - I'll change that now!

Dockerfile
gracewgao: Hmm, I just got rid of the other instances where I specified /bin/bash and just kept this one

gracewgao:
I've made some changes that you suggested - let me know if there's anything else?

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    && opam switch create 4.10.2
+
+# creates symlink for Python 3
+RUN sudo ln -s /usr/bin/python3 /usr/bin/python
+
+# updates PATH variable for opam
+ENV PATH="/home/opam/.opam/4.10.2/bin:$PATH"
+
+# clones pyre-check and installs dependencies
+RUN git clone https://github.com/facebook/pyre-check
```
gbleaney, line 17: We should use `ADD` to grab files from the local filesystem:
https://docs.docker.com/engine/reference/builder/#add

Note that we probably want to do this in three steps, to prevent misc file changes from causing the unzip and pip install steps from running:
1. Pull in zip file (then unzip)
2. Pull in requirements.txt (then pip install)
3. Pull in remaining files

gbleaney (changes requested):
Getting really close! Mostly just docs tweaks at this point

gracewgao:
Just updated the changes you suggested! The ADD command took a bit of permission toggling but the current setup works for me. Regarding this, I also added a quick note in the installation docs that the pyre-check files should only be changed locally and then rebuilt in the Docker image.

gbleaney:
@gracewgao I'm putting this through our internal review system now. In the mean time, can you please add some evidence of testing?

gracewgao:
@gbleaney Sure! Apart from the integration tests and tests run inside the Docker container, would you like me to manually test it and attach screenshots?

documentation/website/docs/installation.md
```diff
@@ -86,6 +86,43 @@ $ echo 'PATH="/path/to/pyre-check/scripts:$PATH"' >> ~/.bashrc
...
+   ```  
+
+2. Build the Docker image with the tag `pyre-check` (or another tag if you wish)
+   ```bash
+   docker build -t pyre-docker .
+   ```   
+
+3. Run the new image in a new container `pyre-container` (or another name if you wish)
+   ```bash
+   docker run \
```
arthaud, line 108: Here and other places: please remove those trailing whitespaces
```suggestion
   docker run \
```

documentation/website/docs/installation.md
```diff
@@ -86,6 +86,43 @@ $ echo 'PATH="/path/to/pyre-check/scripts:$PATH"' >> ~/.bashrc
...
+
+4. Inside the container, run any Pyre command now with `pyre`!
+
+   *Note: When initializing Pyre with `pyre init`, you may have to enter the following paths for the binary and typeshed:*
+   ```bash
+   ƛ No `pyre.bin` found, enter the path manually: /home/opam/pyre-check/source/_build/default/main.exe
+   ƛ Unable to locate typeshed, please enter its root: /home/opam/pyre-check/stubs/typeshed/typeshed-master
+   ```
+
+**For contributors:** Inside the Docker container, the added pyre-check directory is only editable by the root user. To contribute to Pyre, make edits in your local filesystem and rebuild the Docker by running Step 2, then running a new DOcker container in Steps 3-4.
```
arthaud, line 124: Typo
```suggestion
**For contributors:** Inside the Docker container, the added pyre-check directory is only editable by the root user. To contribute to Pyre, make edits in your local filesystem and rebuild the Docker by running Step 2, then running a new Docker container in Steps 3-4.
```

arthaud (changes requested):

gracewgao:
@arthaud Thanks for catching that - I just fixed up the whitespace and typo!

PR #409, opened 2021-04-16
refactored pysa_server.py to stage for Pysa's Language Server specific changes

m0mosenpai:
Expected outcome for this PR is to display and highlight errors for `.pysa` files, via the Pysa VSCode Extension.
This is part of the larger issue of creating a VSCode extension for Pysa in the MLH-Fellowship program listed [here](https://github.com/MLH-Fellowship/pyre-check/issues/4) and has been discussed with @gbleaney before.

Possible milestones:
- [X] Make template extension code for Pysa by cloning Pyre
- [X] Add new functions necessary for handling model validation errors to `pysa_server.py` and `persistent.py`
- [X] Connect everything and call `get_invalid_taint_models()` in `query.py`
- [X] Make relevant changes on the extension side
- [x] Clean up the server of old Pyre code

Output so far:
image

Errors are being calculated but are not being published due to `server_state.opened_documents` list being empty as show in the output. The publishing diagnostics function iterates over this list and sends the errors to VSCode. It's empty right now and hence, the errors are never published. (Opening/closing documents is not being detected)


gbleaney:
> Errors are being calculated but are not being published due to server_state.opened_documents list being empty as show in the output. The publishing diagnostics function iterates over this list and sends the errors to VSCode. It's empty right now and hence, the errors are never published. (Opening/closing documents is not being detected)

It looks like the server state is modified here:
https://github.com/facebook/pyre-check/blob/3a61444e794a0aac6fa3582080f4e1d343cf39a9/client/commands/v2/pysa_server.py#L380

Can you add some debugging statements to see if the `process_open_request` / `process_close_request` functions are being called when you open and close files in the plugin?

Also, as a workaround, can you try just publishing all errors, regardless of open and close status?

m0mosenpai:
> > Errors are being calculated but are not being published due to server_state.opened_documents list being empty as show in the output. The publishing diagnostics function iterates over this list and sends the errors to VSCode. It's empty right now and hence, the errors are never published. (Opening/closing documents is not being detected)
> 
> It looks like the server state is modified here:
> https://github.com/facebook/pyre-check/blob/3a61444e794a0aac6fa3582080f4e1d343cf39a9/client/commands/v2/pysa_server.py#L380
> 
> Can you add some debugging statements to see if the `process_open_request` / `process_close_request` functions are being called when you open and close files in the plugin?
> 
> Also, as a workaround, can you try just publishing all errors, regardless of open and close status?

I don't think the functions are being called. Can we skip open/close ? I mean, the publishing of errors relies on document's path to know which file we are talking about. 
Also, I was wondering if `.pysa` files are even being tracked in the first place. Maybe it doesn't know which files to monitor for open/close status. Anyway I can make sure if that part is working or not?

m0mosenpai:
**Current working outputs:**

image
image
image

**Where it fails/ doesn't work as expected**:

https://github.com/facebook/pyre-check/issues/431

client/commands/v2/persistent.py
```diff
@@ -503,6 +504,36 @@ def type_errors_to_diagnostics(
     return result
 
 
+def invalid_model_to_diagnostic(
```
gbleaney, line 507: These two functions are specific to Pysa, so they probably belong in client/commands/v2/pysa_server.py, rather than here

client/commands/v2/start.py
```diff
@@ -526,7 +526,7 @@ def get_server_identifier(configuration: configuration_module.Configuration) ->
 
 
 def _write_argument_file(output_file: IO[str], arguments: Arguments) -> None:
-    LOG.info(f"Writing server startup configurations into {output_file.name}...")
+    # LOG.info(f"Writing server startup configurations into {output_file.name}...")
```
gbleaney, line 529:
```suggestion
    LOG.info(f"Writing server startup configurations into {output_file.name}...")
```

documentation/pysa_tutorial/exercise2/.pyre_configuration
```diff
@@ -10,5 +10,6 @@
   ],
   "exclude": [
     ".*/integration_test/.*"
-  ]
+  ],
+  "use_command_v2": true
```
gbleaney, line 14: I'm not opposed to committing this, but if we do it should be in a separate PR and should affect all exercises

documentation/pysa_tutorial/exercise2/views.py
```diff
@@ -21,7 +21,7 @@ def operate_on_threes(request: HttpRequest) -> HttpResponse:
 
     exec(f"result = 3 {operator} 3")
 
-    return result  # noqa: F821
+    return result  # noqa: F82 1
```
gbleaney, line 24:
```suggestion
    return result  # noqa: F821
```

tools/ide_plugins/pysa-vscode/src/main.ts
```diff
@@ -39,7 +39,7 @@ export async function activate(_: vscode.ExtensionContext) {
     };
     
     let clientOptions: LanguageClientOptions = {
-        documentSelector: [{scheme: 'file', language: 'python'}],
```
gbleaney, line 42: Given that @gracewgao's changes have landed, you can probably use the `pysa_model` language from #426

client/commands/v2/pysa_server.py
```diff
@@ -4,81 +4,78 @@
 # LICENSE file in the root directory of this source tree.
 
 
-"""
-This is a temporary implementation of Pysa's language server. This is an exact copy
```
gbleaney, line 8: Docs are still good to have. Can we update rather than delete?

client/commands/v2/pysa_server.py
```diff
@@ -4,81 +4,78 @@
...
         self.server_identifier = server_identifier
-        self.pyre_arguments = pyre_arguments
-        self.client_output_channel = client_output_channel
-        self.server_state = server_state
+
+    async def update_errors(self) -> None:
+        pyre_connection = api_connection.PyreConnection(
+            Path(self.pyre_arguments.global_root)
+        )
+        model_errors = query.get_invalid_taint_models(pyre_connection)
```
gbleaney, line 70: We needed a try/catch here, right?

client/commands/v2/pysa_server.py
```diff
@@ -100,231 +97,11 @@ async def log_and_show_message_to_client(
             LOG.debug(message)
         await self.show_message_to_client(message, level)
 
-    def update_type_errors(self, type_errors: Sequence[error.Error]) -> None:
```
gbleaney, line 103: I love the red :D

client/commands/v2/pysa_server.py
```diff
@@ -100,231 +97,11 @@ async def log_and_show_message_to_client(
...
-        pyre_manager: connection.BackgroundTaskManager,
-    ) -> None:
-        self.input_channel = input_channel
-        self.output_channel = output_channel
-        self.client_capabilities = client_capabilities
-        self.state = state
-        self.pyre_manager = pyre_manager
+    async def show_model_errors_to_client(self, diagnostics) -> None:
+        for path, diagnostic in diagnostics.items():
+            await _publish_diagnostics(self.output_channel, path, [])
```
gbleaney, line 102: Why do we have two calls to `_publish_diagnostics`?

client/commands/v2/pysa_server.py
```diff
@@ -357,19 +123,9 @@ async def process_open_request(
...
-        if not self.pyre_manager.is_task_running():
-            await self._try_restart_pyre_server()
-        else:
-            document_diagnostics = self.state.diagnostics.get(document_path, None)
-            if document_diagnostics is not None:
-                LOG.info(f"Update diagnostics for {document_path}")
-                await _publish_diagnostics(
-                    self.output_channel, document_path, document_diagnostics
-                )
+        await _publish_diagnostics(self.output_channel, document_path, [])
```
gbleaney, line 127: Why have these two functions called in sequence? It seems like it'd be cleaner to just have one function to call

client/commands/v2/pysa_server.py
```diff
@@ -491,23 +230,15 @@ async def run_persistent(
...
-                        client_output_channel=stdout,
-                        server_state=initial_server_state,
-                    )
-                ),
+                binary_location=binary_location,
+                server_identifier=server_identifier,
+                pyre_arguments=pysa_arguments,
             )
-            return await server.run()
+            return await server._run()
```
gbleaney, line 241: `_` methods are considered private in Python. If we intend to call it from outside the class, the method should stay as `run` rather than `_run`

gbleaney (changes requested):
Giving a preliminary review here, pending the split into multiple diffs that we discussed yesterday

client/commands/v2/pysa_server.py
m0mosenpai: There was an issue where the errors after publishing, would not clear up automatically once I corrected them. (The red squiggly line would not go away). I was missing a call to publish function with empty arguments (which clears the errors), which was there originally but got lost during the refactoring. I suppose it'll be better to add this call inside `update_errors()` itself? Just one function to call then.

client/commands/v2/pysa_server.py
m0mosenpai: That is by mistake, my bad. I was trying some stuff out.

client/commands/v2/pysa_server.py
```diff
@@ -5,80 +5,71 @@
 
 
 """
-This is a temporary implementation of Pysa's language server. This is an exact copy
-of Pyre's language server, which lives in persistent.py.
+This is a tempororary implementation of Pysa's language server. It is a refactored
```
gbleaney, line 8:
```suggestion
This is an implementation of Pysa's language server. It is a refactored
```

client/commands/v2/pysa_server.py
```diff
@@ -5,80 +5,71 @@
...
+version of persistent.py, preparing for upcoming Pysa specific changes.
```
gbleaney, line 9:
```suggestion
version of persistent.py.
```

client/commands/v2/pysa_server.py
```diff
@@ -100,231 +91,11 @@ async def log_and_show_message_to_client(
...
-        client_capabilities: lsp.ClientCapabilities,
-        state: ServerState,
-        pyre_manager: connection.BackgroundTaskManager,
-    ) -> None:
-        self.input_channel = input_channel
-        self.output_channel = output_channel
-        self.client_capabilities = client_capabilities
-        self.state = state
-        self.pyre_manager = pyre_manager
+    async def show_model_errors_to_client(self, diagnostics) -> None:
```
gbleaney, line 94: It looks like this function snuck in, and should probably be in the next PR, right?

client/commands/v2/pysa_server.py
```diff
@@ -5,80 +5,71 @@
...
+class PysaServer:
+    # I/O Channels
+    input_channel: connection.TextReader
+    output_channel: connection.TextWriter
+
+    # Immutable States
+    client_capabilities: lsp.ClientCapabilities
+
+    # Mutable States
+    state: ServerState
```
arthaud, line 50: It looks like our linter is not happy because this is never used. Could you delete it?

arthaud (changes requested):

Issue #411, opened 2021-04-19
Pyre returns different results with the same input for sqlite3 (all SQL injections found) and mysql.connector (just 1 injection found out of 3)


maximmasiutin:
I have created a taint sink file for mysql.connector (mysql_connector_sinks.pysa) and a sample application (unsafe-mysql.py), with 3 subsequent lines, each of them sends an SQL-injectable query to the server. However, pyre only detects the last string as vulnerable, and if I comment it, it detects none. However, if I send the same strings to sqlite3 (using sqlite3_sinks.pisa that ships with pyre-check), it detects all three. I have attached all the files to reproduce the issue. How can I debug pyre to figure out why I obtain different results with the same input for sqlite3 and mysql.connector? Can you help? Or is there a Pyre forum or user discussion group where I may ask this question?

[mysql-connector-sinks.zip](https://github.com/facebook/pyre-check/files/6336412/mysql-connector-sinks.zip)

I have the following files in the "taint_models_path" directory:
```
collection_propagation.pysa
filesystem_other_sinks.pysa
filesystem_sinks.pysa
format_string_sinks.pysa
general.pysa
http_server.pysa
input.pysa
logging_sinks.pysa
mysql_connector_sinks.pysa
protocols.pysa
rce_sinks.pysa
requests_api_sinks.pysa
sanitizers.pysa
skipped_overrides.pysa
sqlite3_sinks.pysa
taint.config
xss_sinks.pysa

```

onionymous:
Hello @maximmasiutin! The reason you can't detect issues with your taint models for `mysql.connector` is because your project does not have type information for the `mysql` module. Our documentation at https://pyre-check.org/docs/pysa-false-negatives#missing-type-information describes how to debug these false negatives. Adding `reveal_taint()` and `reveal_type()` calls to your code makes this clear. 

For unsafe-sqlite.py, this is the output:
```
2021-04-19 11:06:42,980 [PID 16629] WARNING unsafe-sqlite:5:0-5:11: Revealed type for db: sqlite3.dbapi2.Connection
2021-04-19 11:06:42,980 [PID 16629] WARNING unsafe-sqlite:7:0-7:11: Revealed type for cur: sqlite3.dbapi2.Cursor
[...]

Issue #413, opened 2021-04-19
Update Django .pyi files


maximmasiutin:
I have made a pull request to update the Django .pysa file "django_sources_sinks.pysa" to the current release of Django, since the existing .pysa file gives errors preventing "pyre" from running, see https://github.com/facebook/pyre-check/pull/412

However, there are lots of Django .pyi files at https://github.com/facebook/pyre-check/tree/master/stubs/django made by Facebook which also need to be updated.

What about an automated tool like from "stubgen" from "mypy", and then write a script that modifies .pyi files to insert types instead of "Any", but only those lines that are relevant to "django_sources_sinks.pysa".

But a better idea is to contribute to https://github.com/django/django to insert types in the Django code directly.

onionymous:
So this is part of a bigger discussion that we're having internally, but the **tl;dr:** these type stubs are no longer actively being maintained. In general the focus of Pyre/Pysa is just typechecking and static analysis, and we'd prefer not to maintain type stubs if possible - it's best if existing projects having their own type information, or we rely on typing projects like Typeshed. Since there exists https://github.com/typeddjango/django-stubs, which seems like a far more actively maintained and complete set of type stubs for Django, we're also thinking of switching to use this.

In the meantime, to unblock your efforts to run Pysa on Django projects, you might consider deleting Pyre's folder for Django type stubs completely, and trying to use the stubs at https://github.com/typeddjango/django-stubs instead.

We will discuss internally about what the best way to move forward is, but for now we aren't expecting to update these stubs.

maximmasiutin:
Thank you, I will take the https://github.com/typeddjango/django-stubs file. However, an updated .pysa file for Django is still needed, because in the latest version of Django, "pyra analyze" gives errors unless you update the "django_sources_sinks.pysa" file.

onionymous:
Yes, that is true. Thank you for the PR - we'll get around to looking at the changes in #412 as soon as possible.

maximmasiutin:
I'm still not yet very comfortable with testing Django applications with "pyre analyze". Once I manage to configure Pyre to find vulnerabilities in the deliberately vulnerable Django applications, and we will be able to make sure that the https://github.com/typeddjango/django-stubs work with Pyre properly, we should probably delete the folder with "django" .pyi files from the Pyre repository, and instead add a comment to "django_sources_sinks.pysa" and/or somewhere else on how to get the https://github.com/typeddjango/django-stubs and where to put them in order everything worked properly.

[...]

PR #414, opened 2021-04-19
Sinks and type definitions for the mysql.connector

maximmasiutin:
Here are the sinks and the type definitions for the mysql.connector. With these files, Pyre is able to find all three SQL injection vulnerabilities in the following example:
[unsafe-mysql-py.zip](https://github.com/facebook/pyre-check/files/6338706/unsafe-mysql-py.zip)

Fixes https://github.com/facebook/pyre-check/issues/411


arthaud:
That's awesome!
cc @onionymous for the .pysa
Could you make a pull request to https://github.com/python/typeshed with the mysql stubs? This way, everyone could benefit from them, not just pyre.

onionymous:
Thank you very much for doing this! The Pysa models look good to me and are in the correct location.

+1 for the Typeshed PR - the type stubs look amazing!

maximmasiutin:
These mysql stubs are very basic, I've trimmed them to just be useful for Pyre/Pysa, and deleted the rest. 

In order to push them to https://github.com/python/typeshed -- they need to be more or less complete.

Also, there is a message about MySQL stubs at https://github.com/python/typeshed/issues/146#issuecomment-680783332

onionymous:
Perhaps you can consider submitting a PR to add typing to the mysql project itself (https://github.com/mysql/mysql-connector-python)? I believe it would be a smaller change than this PR, since you'd just be adding annotations for the connection and cursor objects. As with the discussion on #413, generally we'd prefer not to maintain type stubs if possible.

However, if there are further issues, we would be happy to merge this, since it's a really great contribution that improves our coverage.

maximmasiutin:
@onionymous Thank you! Please consider merging this request, since we still need the .pysa file, even if the developers of the mysql-connector-python will add the annotations for the connection and cursor object. Then, we will be able to delete the .pyi files for MySQL from the pyre-check repository.

maximmasiutin:
Please consider accepting this pull request. When and if the developers of the "mysql.connector" will add the stubs, we may delete the stubs that I have prepared from the repository. These stubs are mainly focused on SQL Injections for Pyre, so they are incomplete. They only define the connection and cursor types for Pyre. 

Most probably, the developers of the "mysql.connector" may need the stubs fully converted. I have submitted a https://github.com/mysql/mysql-connector-python/pull/67 , though.

maximmasiutin:
If you accept this pull request, please also add the stubs/mysql directory to the list of directories checked by GitHub on pull request, e.g. on the validity of code formatting by "black".

gbleaney:
@maximmasiutin thanks for putting up https://github.com/mysql/mysql-connector-python/pull/67! I think that you could actually get away with a much smaller PR. I think @onionymous was suggesting you add explicit inline return types. If you add those types in these places, I think that would be enough to give Pysa the type information it needs to catch the SQL vulnerabilities you had in #411:

https://github.com/mysql/mysql-connector-python/blob/d1e89fd3d7391084cdf35b0806cb5d2a4b413654/lib/mysql/connector/__init__.py#L188

https://github.com/mysql/mysql-connector-python/blob/86cceb2bc914362e0990459002976c7e5a271e1d/lib/mysql/connector/abstracts.py#L992

https://github.com/mysql/mysql-connector-python/blob/86cceb2bc914362e0990459002976c7e5a271e1d/lib/mysql/connector/abstracts.py#L1054

That solution is actually preferable to the `.pyi` file solution, because the presence of `.pyi` files means Pyre and Pysa are not allowed to look at the source code of the files those `.pyi` files add typeing too ([see the PEP](https://www.python.org/dev/peps/pep-0484/#stub-files)).

> Please consider accepting this pull request.

Let's wait a bit to see the result of https://github.com/mysql/mysql-connector-python/pull/67. If we can't get that landed in a reasonable time, we'll definitely land this request because it's adding some really awesome SQL sink coverage that will benefit you, us, and probably a lot of other open source users.

maximmasiutin:
@gbleaney Thank you! I knew that @onionymous was suggesting to modify the .py files to just add types in a few places, but I don't think that the developers of mysql-connector would have allowed this. It would have broken compatibility with Python2.

gbleaney:
@maximmasiutin there is actually a backwards compatible way to add types to Python code:
https://www.python.org/dev/peps/pep-0484/#type-comments

It's more limited that what's available in later versions of Python 3, but it's more than enough to let you specify function return types

maximmasiutin:
> @maximmasiutin there is actually a backwards compatible way to add types to Python code:
> https://www.python.org/dev/peps/pep-0484/#type-comments

@gbleaney -- Thank you! I'm aware about comment-style type hints for Python 2, but I was not aware that Pyre supports such kind of type hints. Let us wait for the decision on https://github.com/mysql/mysql-connector-python/pull/67 -- if it will be rejected, I will try to ask to add type hint comments by a pull request.

maximmasiutin:
Can you accept this pull request now, and when (and if) the files become available via the 'pip update mysql-connector', you will just delete the .pyi files from the pyre-check repository and only leave the .pysa file?

The developers of MySQL moved the pull request mysql/mysql-connector-python#67  to http://bugs.mysql.com/bug.php?id=103488 and closed the pull request on GitHub.

Anyway, it takes time between accepting the request and the availability of files via the 'pip install'.

gbleaney:
@maximmasiutin I've imported your request to our internal CI so I can see if it creates any issues. At the same time, I'll check with the team to see if we're OK merging as you've proposed. I really don't want us maintaining these stubs, but the models do add some very valuable coverage.

maximmasiutin:
@gbleaney 
Thank you! I see there are issues with the internal linter. Can you let me know the exact details, so I fix them?

gbleaney:
@maximmasiutin the linter is complaining about two things:

1. Every file needs a copyright header:
```
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.
```
2. The [black](https://github.com/psf/black) formatter needs to be run.

maximmasiutin:
OK, Thank you! I was using Black from pip install, but did only support .pyi and .pi file, not the .pysa files. Will check that one from https://github.com/psf/black, not from the pip install.

maximmasiutin:
I have added the copyright headers to the .pyi files. As about the Black, the files were already formatted.

gbleaney:
@maximmasiutin Strange, you must be on a different version of Black. Internally, it's complaining that it wants the `from typing import Any` imports at the top, above the relative imports. Anyway, I can just run the auto-fixer for these issues before I land it.

maximmasiutin:
@gbleaney - I installed it by pip3 install git+git://github.com/psf/black 

When I run `black --version` I got `black, version 21.4b2`

I run just `black .` without any command-line parameters.

maximmasiutin:
When I run black, it does not give any error, but when I run PyLint, I get errors.

For example:

`pylint abstracts.pyi | grep ": E"`


```
abstracts.pyi:7:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
abstracts.pyi:8:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
abstracts.pyi:9:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
abstracts.pyi:14:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
abstracts.pyi:15:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
abstracts.pyi:32:26: E0602: Undefined variable 'MySQLConnectionAbstract' (undefined-variable)
abstracts.pyi:54:40: E0601: Using variable 'MySQLCursorAbstract' before assignment (used-before-assignment)
```

maximmasiutin:
If I go to lines 32 and 54 (where the linter gives errors), I see the following:

32:
`    
def database(self) -> MySQLConnectionAbstract: ...
`

54:
`    
def info_query(self, query: Any) -> MySQLCursorAbstract: ...
`

gbleaney:
@maximmasiutin I believe Pyre supports postponed evaluation of annotations, but if you want to make PyLint happy, you can put the `MySQLCursorAbstract` annotations in quotes. More details here:
https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class

As for the "Attempted relative import" errors, it looks like you've got import statements for non-existant modules. To work around that, you can remove the imports or use the `__getattr__` trick described here in the "Stub files may be incomplete." bullet:
https://www.python.org/dev/peps/pep-0484/#stub-files

maximmasiutin:
@gbleaney - I've fixed the lint errors. Please review. By the way, I had to install the mysql-connector module via "pip install" in order to avoid the pylint errors about missing modules.

PR #420, opened 2021-04-24
fix(docs): corrects common typos in project documentation

plan-do-break-fix:
Scope of changes is restricted to docs only.
Corrections made in agreement with Standard American English.


gbleaney:
Thanks for fixing these! I've imported the fixes and they'll land as soon as someone else approves (likely not until tomorrow)

PR #426, opened 2021-05-11
Set up Pysa syntax highlighting

gracewgao:
The goal here is to highlight `taint.config` files as regular JSON files and highlight `.pysa` files using Python grammar rules with a couple of modifications.

To do this, I declared new languages for both file types in `package.json`, and made a TextMate file for each. For the `taint.config`, it includes the `source.json` scope for the entire file and inherits all the grammar rules so that it's essentially treated the same as a JSON file. For the Pysa files, I initially tried to include the Python grammar rules and define new ones on top of that, but there were a lot of conflicts between the scopes. It also seemed redundant to undefine already-defined rules and I think forking the rules directly would also allow for more flexible customization in the future.

So, I forked the official Python copy from [here](https://github.com/textmate/python.tmbundle/blob/master/Syntaxes/Python.tmLanguage) and made the following changes:
- Highlighting feature annotations mentioned in these [Pysa docs](https://pyre-check.org/docs/pysa-features/) - `TaintSource`, `TaintSink`, `TaintInTaintOut`, `Sanitize`, `Via`, `AttachToSource`, `AttachToSink`, `AttachToTito`, `AddFeatureToArgument`, `ViaValueOf`, `WithTag`, `ViaDynamicFeature` as part of the `support.class.pysa` scope
- Highlighting fully-qualified function names as part of the `entity.name.function.pysa` scope
- Removing `self`, `type`, `__init__` from the Python grammar rules

For more information about naming conventions of scopes, check the TextMate [scope docs](https://macromates.com/manual/en/language_grammars).

Here are some screenshots from running the extension in my VSCode environment:

Highlighting for feature annotations:

Highlighting for fully-qualified functions

Highlighting for fully-qualified functions:

Highlighting for feature annotations

Highlighting for `taint.config` files:

Highlighting for taint.config files

Let me know if I missed anything or if you have any other ideas for syntax highlighting!


tools/ide_plugins/pysa-vscode/language-configuration.json
```diff
@@ -0,0 +1,30 @@
+{
+    "comments": {
+        // symbol used for single line comment. Remove this entry if your language does not support line comments
+        "lineComment": "//",
```
gbleaney, line 4:
```suggestion
        "lineComment": "#",
```

tools/ide_plugins/pysa-vscode/language-configuration.json
```diff
@@ -0,0 +1,30 @@
...
+        // symbols used for start and end a block comment. Remove this entry if your language does not support block comments
+        "blockComment": [ "/*", "*/" ]
```
gbleaney, line 6: This should probably be deleted

tools/ide_plugins/pysa-vscode/language-configuration.json
```diff
@@ -0,0 +1,30 @@
...
+    ],
+    // symbols that are auto closed when typing
+    "autoClosingPairs": [
+        ["{", "}"],
+        ["[", "]"],
+        ["(", ")"],
+        ["\"", "\""],
+        ["'", "'"]
+    ],
+    // symbols that can be used to surround a selection
```
gbleaney, line 22: Should triple quotes be included in this section?

tools/ide_plugins/pysa-vscode/package.json
```diff
@@ -38,5 +38,34 @@
...
+  "contributes": {
+      "languages": [{
+          "id": "pysa",
+          "aliases": ["Pysa", "pysa"],
+          "extensions": [".pysa"],
+          "configuration": "./language-configuration.json"
+      },
+      {
+          "id": "taint",
+          "aliases": ["Pysa", "taint"],
```
gbleaney, line 54: Is it OK for two languages to share the same alias? What is the alias used for?

tools/ide_plugins/pysa-vscode/package.json
```diff
@@ -38,5 +38,34 @@
...
+      }
+      ],
+      "grammars": [{
+          "language": "pysa",
+          "scopeName": "source.pysa",
+          "path": "./syntaxes/Pysa.tmLanguage"
+      },
+      {
+          "language": "taint",
+          "scopeName": "taint.config",
```
gbleaney, line 66: Naming is inconsistent between these two languages. Should this be `source.taint`?

tools/ide_plugins/pysa-vscode/package.json
```diff
@@ -38,5 +38,34 @@
     "tslint-microsoft-contrib": "^5.0.1",
     "typescript": "^3.5.1",
     "vscode": "^1.1.7"
+  },
+  "categories": [
+    "Programming Languages"
+  ],
+  "contributes": {
+      "languages": [{
```
gbleaney, line 46: Nit: The naming of the two languages seems to put `.pysa` files above `taint.config` files, since `.pysa` files are the ones that get put into the "pysa" language. What if we put these on the same level, perhaps naming one `pysa_models` and the other `pysa_config`, or something like that?

tools/ide_plugins/pysa-vscode/syntaxes/Pysa.tmLanguage
```diff
@@ -0,0 +1,4007 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gbleaney, line 1: A few requests here:
1. Can you provide the source so we can check licensing?
2. Can you point out which lines you changed? Maybe the output of `diff` or something? It'll be too cumbersome to review a whole fork of a 3k line file.

tools/ide_plugins/pysa-vscode/syntaxes/Pysa.tmLanguage
```diff
@@ -0,0 +1,4007 @@
...
+					<key>include</key>
+					<string>#string_quoted_single</string>
+				</dict>
+			</array>
+		</dict>
+	</dict>
+	<key>scopeName</key>
+	<string>source.pysa</string>
+	<key>uuid</key>
+	<string>F23DB5B2-7D08-11D9-A709-000D93B6E43C</string>
```
gbleaney, line 4005: Does this UUID need to be regenerated? I have no idea where it's used, but I can imagine two files having the same UUID is a bad idea

tools/ide_plugins/pysa-vscode/package.json
```diff
@@ -38,5 +38,34 @@
...
+          "id": "pysa",
+          "aliases": ["Pysa", "pysa"],
+          "extensions": [".pysa"],
+          "configuration": "./language-configuration.json"
+      },
+      {
+          "id": "taint",
+          "aliases": ["Pysa", "taint"],
+          "extensions": ["taint.config"],
+          "configuration": "./language-configuration.json"
```
gbleaney, line 56: Should we have a separate language configuration here? JSON has some different rules from Python

gbleaney (changes requested):

tools/ide_plugins/pysa-vscode/package.json
```diff
@@ -38,5 +38,34 @@
...
+      }
+      ],
+      "grammars": [{
```
arthaud, line 59: Formatting is pretty inconsistent within that file. In my opinion, we should break a line here and indent the following object. Could we use an auto formatter here?

arthaud (changes requested):

tools/ide_plugins/pysa-vscode/package.json
gracewgao: I've made separate configuration files `model-language-configuration.json` for `.pysa` files and `config-language-configuration.json` for `taint.config` files using the official VSCode configurations for [JSON](https://github.com/microsoft/vscode/blob/main/extensions/json/language-configuration.json) and [Python](https://github.com/microsoft/vscode/blob/main/extensions/python/language-configuration.json)

tools/ide_plugins/pysa-vscode/package.json
gracewgao: Thanks for catching that, I just used Prettier to fix it - let me know if it looks better now!

tools/ide_plugins/pysa-vscode/package.json
gracewgao: I changed both scope names to `source.pysa.model` and `source.pysa.config` to keep it consistent with the language-naming - what do you think?

tools/ide_plugins/pysa-vscode/package.json
gracewgao: I like that idea! I just made those changes and reflected them in the file and scope naming as well.

tools/ide_plugins/pysa-vscode/syntaxes/Pysa.tmLanguage
gracewgao: You're right - I just generated a new UUID for each language
image

tools/ide_plugins/pysa-vscode/language-configuration.json
gracewgao: I've replaced both language configuration files with the Python and JSON copy from [here](https://github.com/microsoft/vscode/tree/main/extensions)

tools/ide_plugins/pysa-vscode/syntaxes/Pysa.tmLanguage
gracewgao: Yep, for sure - I've made a gist [here](https://gist.github.com/gracewgao/f816b7c27ad66b76df9961e3e43a0d19/revisions) to compare the differences. To summarize, I basically just changed the scope names to be `pysa` rather than `python`, fixed the UUID, and made all the other changes I mentioned in the PR description.

tools/ide_plugins/pysa-vscode/package.json
gracewgao: Honestly, I'm not 100% sure but as far as I can tell, this would change the way the language is displayed in VSCode:

image

I did some digging around and it seems it might possibly cause problems like the one mentioned [here](https://github.com/microsoft/vscode-docs/issues/2862), so I just changed the aliases to be `Pysa Model` and `Pysa Config`

tools/ide_plugins/pysa-vscode/config-language-configuration.json
```diff
@@ -0,0 +1,19 @@
+// JSON language-configuration taken from https://github.com/microsoft/vscode/blob/main/extensions/json/language-configuration.json
```
gbleaney, line 1: Is there a way to include this file and the other `.json` file, rather than straight up copy-pasting them? Eg. something like what you did in `PysaConfig.tmLanguage` to just include the JSON grammar, rather than copy it in.

tools/ide_plugins/pysa-vscode/syntaxes/PysaModel.tmLanguage
```diff
@@ -0,0 +1,4007 @@
...
+		<string>^#!/.*\bpysa[\d\.]*\b</string>
+		<key>keyEquivalent</key>
+		<string>^~P</string>
+		<key>name</key>
+		<string>Pysa</string>
+		<key>patterns</key>
+		<array>
+			<dict>
+				<key>comment</key>
+				<string>feature annotations in Pysa</string>
```
gbleaney, line 19:
```suggestion
				<string>Taint and Feature keywords in Pysa</string>
```

tools/ide_plugins/pysa-vscode/syntaxes/PysaModel.tmLanguage
```diff
@@ -0,0 +1,4007 @@
...
+				<key>match</key>
+				<string>\b(TaintSource|TaintSink|TaintInTaintOut|Sanitize|Via|AttachToSource|AttachToSink|AttachToTito|AddFeatureToArgument|ViaValueOf|WithTag|ViaDynamicFeature)\b</string>
```
gbleaney, line 21: Don't forget:
- `AppliesTo`
- `ViaTypeOf`

tools/ide_plugins/pysa-vscode/syntaxes/PysaModel.tmLanguage
```diff
@@ -0,0 +1,4007 @@
...
+						str | sum | super | tuple | unichr | unicode | vars |
+						xrange | zip)
+					\b\s*(?!(\())</string>
+				<key>name</key>
+				<string>support.function.builtin.name.pysa</string>
+			</dict>
+			<key>builtin_types</key>
+			<dict>
+				<key>comment</key>
+				<string>These are from https://docs.pysa.org/X/library/stdtypes.html
```
gbleaney, line 1504:
```suggestion
				<string>These are from https://docs.python.org/X/library/stdtypes.html
```

tools/ide_plugins/pysa-vscode/syntaxes/PysaModel.tmLanguage
```diff
@@ -0,0 +1,4007 @@
...
+					<key>2</key>
+					<dict>
+						<key>name</key>
+						<string>support.function.magic.name.pysa</string>
+					</dict>
+				</dict>
+				<key>comment</key>
+				<string>These methods have magic interpretation by python and are generally
+					   called indirectly through syntactic constructs. Names are from
+					   https://docs.pysa.org/X/reference/datamodel.html where X is 2.7 and 3.5
```
gbleaney, line 2069:
```suggestion
					   https://docs.python.org/X/reference/datamodel.html where X is 2.7 and 3.5
```

gbleaney (changes requested):
Looking pretty good! Just a few more changes and then I think we're good to land

tools/ide_plugins/pysa-vscode/config-language-configuration.json
gracewgao: Looks like this is an issue that's been brought up and is [currently in the backlog](https://github.com/microsoft/vscode/issues/2871) - I don't think there's any official VSCode support for it, but I can try the workarounds mentioned if we really don't want to copy and paste. Since language configuration is used for autocomplete etc. within the IDE, I don't think there's any way to go around VSCode to get it :/

gbleaney:
LGTM now. Thanks for the quick turnaround. It'll go through our internal review process now, so there may be a couple more comments, but we're on the path to landing!

PR #427, opened 2021-05-20
[pyre] script to convert output to SARIF

dkgi:


client/pyre.py
```diff
@@ -37,6 +37,14 @@
 LOG: logging.Logger = logging.getLogger(__name__)
 
 
+def one() -> int:
```
shannonzhu, line 40: testing leftovers?

dkgi: Yeah, it's still not working. I'm trying to get results to show up in this PR.

PR #433, opened 2021-06-01
added support for Pysa Model Validation in VSCode Extension

m0mosenpai:
Follow up to https://github.com/facebook/pyre-check/pull/409

This PR adds the implementation of model validation feature into Pysa's VSCode Plugin.

### Working  Examples

Screenshots showing the running extensions after these changes (the errors displayed are such due to misconfigured typeshed)

image

The edge case where Pyre is not able to parse the file, the following error is displayed and the extension continues to work:
image


client/commands/v2/pysa_server.py
```diff
@@ -5,80 +5,116 @@
...
+    ) -> Dict[Path, List[lsp.Diagnostic]]:
+        result: Dict[Path, List[lsp.Diagnostic]] = {}
+        for model in invalid_models:
+            result.setdefault(Path(model.path), []).append(
+                self.invalid_model_to_diagnostic(model)
+            )
+        return result
+
+    async def update_errors(self, document_path: Path) -> None:
+        await _publish_diagnostics(self.output_channel, document_path, [])
```
gbleaney, line 100: What is the purpose of this line? Is it just to overwrite the existing set of errors? Won't that happen later anyway?

client/commands/v2/pysa_server.py
```diff
@@ -100,231 +136,13 @@ async def log_and_show_message_to_client(
...
-        state: ServerState,
-        pyre_manager: connection.BackgroundTaskManager,
-    ) -> None:
-        self.input_channel = input_channel
-        self.output_channel = output_channel
-        self.client_capabilities = client_capabilities
-        self.state = state
-        self.pyre_manager = pyre_manager
+        for path, diagnostic in diagnostics.items():
+            await _publish_diagnostics(self.output_channel, path, [])
```
gbleaney, line 143: Couldn't we delete the next two lines and just do this?
```suggestion
            await _publish_diagnostics(self.output_channel, path, diagnostic or [])
```

client/commands/v2/pysa_server.py
```diff
@@ -5,80 +5,116 @@
...
+            source="Pysa",
+        )
+
+    def invalid_models_to_diagnostics(
+        self,
+        invalid_models: Sequence[query.InvalidModel],
+    ) -> Dict[Path, List[lsp.Diagnostic]]:
+        result: Dict[Path, List[lsp.Diagnostic]] = {}
+        for model in invalid_models:
+            result.setdefault(Path(model.path), []).append(
```
gbleaney, line 94: As discussed on our call, you'll have to handle the case of a `None` path

gbleaney (changes requested):
Changes looking pretty good! Can you please rebase on master, now that we've landed your previous PR?

client/commands/v2/pysa_server.py
m0mosenpai: It's to clear out the errors. Without this, it works fine for cases where the file intially doesn't have any errors and then if we introduce some. But in the cases where errors are removed from the file, the red lines and problems in the vscode output don't go away.
Let me know if there's a better way to do this cause I think it used to be fine before the refactoring but afterwards, this is how i managed to make it work.

client/commands/v2/pysa_server.py
m0mosenpai: Oh yeah, just changed and tested it, it works like this

client/commands/v2/pysa_server.py
m0mosenpai: I added an if condition. Let me know if there's a more appropriate warning I should display.

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,56 @@ def __init__(
...
+
+    def invalid_models_to_diagnostics(
+        self,
+        invalid_models: Sequence[query.InvalidModel],
+    ) -> Dict[Path, List[lsp.Diagnostic]]:
+        result: Dict[Path, List[lsp.Diagnostic]] = {}
+        for model in invalid_models:
+            if model.path is None:
+                self.log_and_show_message_to_client(
+                    "Path cannot be None", lsp.MessageType.WARNING
```
gbleaney, line 96: We should give the use the error message, rather than something generic like this. General rule: if you're not sure what's going on, give the user enough info so they can figure it out temselves.

client/commands/v2/pysa_server.py
gbleaney: It feel a bit out of place, but I can live with leaving it. But we should at least have a comment that indicates why we have this mysterious call.

gbleaney (changes requested):

client/commands/v2/pysa_server.py
```diff
@@ -28,10 +29,14 @@
     _read_lsp_request,
     try_initialize,
     _log_lsp_event,
+    _publish_diagnostics,
     InitializationExit,
     InitializationSuccess,
     InitializationFailure,
 )
+from api import query, connection as api_connection
+from api.connection import PyreQueryError
```
gbleaney, line 38: These need to be relative imports to work with our internal system

client/commands/v2/pysa_server.py
```diff
@@ -9,6 +9,7 @@
 version of persistent.py.
 """
 
+from pathlib import Path
```
gbleaney, line 12: Don't forget to run black to sort the imports. End result should look something like this:
```
import asyncio
import logging

from pathlib import Path
from typing import List, Sequence, Dict

<REMAINING IMPORTS>
```

client/commands/v2/pysa_server.py
m0mosenpai: I initially tried relative importing, but i get an`ImportError: attempted relative import beyond top-level package` for
```
from ....api import query, connection as api_connection
```

client/commands/v2/pysa_server.py
m0mosenpai: I manually sorted them in a better way. I ran black before but the result was unchanged.

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,56 @@ def __init__(
         self.binary_location = binary_location
         self.server_identifier = server_identifier
 
+    def invalid_model_to_diagnostic(
+        self,
+        invalid_model: query.InvalidModel,
```
arthaud, line 70: pyre complains that `query.InvalidModel` does not exist. Any idea why?

client/commands/v2/pysa_server.py
arthaud: hmm, I'm wondering how we distribute the `api` module. It feels like we should put it under a `pyre` module or something.

arthaud (changes requested):

client/commands/v2/pysa_server.py
m0mosenpai: That seems strange. I'm not sure why it says that.

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,57 @@ def __init__(
...
+
+    def invalid_models_to_diagnostics(
+        self,
+        invalid_models: Sequence[query.InvalidModel],
+    ) -> Dict[Path, List[lsp.Diagnostic]]:
+        result: Dict[Path, List[lsp.Diagnostic]] = {}
+        for model in invalid_models:
+            if model.path is None:
+                self.log_and_show_message_to_client(
+                    f"{model.full_error_message}", lsp.MessageType.WARNING
```
gbleaney, line 96: The messages on line 83 are categorized as errors, but this here is a WARNING. Should we be consistent and make both errors?

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,57 @@ def __init__(
...
+                )
+            else:
+                result.setdefault(Path(model.path), []).append(
```
gbleaney, line 99: Nit: Slightly more common way of doing this is with `defaultdict` (scroll to "Using List as default_factory"):
https://www.geeksforgeeks.org/defaultdict-in-python/

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,57 @@ def __init__(
...
+                    self.invalid_model_to_diagnostic(model)
+                )
+        return result
+
+    async def update_errors(self, document_path: Path) -> None:
+        # Publishing empty diagnostics to clear errors in VSCode
+        await _publish_diagnostics(self.output_channel, document_path, [])
+        pyre_connection = api_connection.PyreConnection(
```
gbleaney, line 107: Might be worth storing this as a server attribute

gbleaney (commented):
Looks pretty great! Some really small comments. At the same time I'll import this and make sure it passes our internal CI

client/commands/v2/pysa_server.py
```diff
@@ -59,6 +65,56 @@ def __init__(
...
+                self.log_and_show_message_to_client(
+                    f"{model.full_error_message}", lsp.MessageType.ERROR
+                )
+            else:
+                result[Path(model.path)].append(self.invalid_model_to_diagnostic(model))
+        return result
+
+    async def update_errors(self, document_path: Path) -> None:
+        # Publishing empty diagnostics to clear errors in VSCode
+        await _publish_diagnostics(self.output_channel, document_path, [])
```
gbleaney, line 108: Feedback from @grievejia:

> If I understand correctly, update_errors will just query for invalid model errors in all files, regardless of whether the currently opened/saved file (i.e. document_path) is? If that's the case I think it's not enough to just clear out diagnostics in document_path. One needs to basically clean up all files in which there are pre-existing diagnostics (you'd probably need to keep track of more states in the server for that purpose).
> 
> Alternatively I wonder if it's possible to either just query invalid model errors for one file. Also, does invalid model error support incremental changes? Is it necessary to query these errors on file open, given that opening a file won't change any of these errors?

gbleaney: This is a good point. Right now we don't have support for incremental or per-file model errors. We just get a list of all possible model errors. I think the simplest solution is just to keep a list of errors we've published previously, and then zero them out when we publish new errors. Thoughts on that solution @grievejia ?

client/commands/v2/pysa_server.py
```diff
@@ -28,10 +31,13 @@
     _read_lsp_request,
     try_initialize,
     _log_lsp_event,
+    _publish_diagnostics,
     InitializationExit,
     InitializationSuccess,
     InitializationFailure,
 )
+from ....api import query, connection as api_connection
```
gbleaney, line 39: These two should be sorted to be just above line 18

gbleaney (changes requested):

client/commands/v2/pysa_server.py
grievejia: > I think the simplest solution is just to keep a list of errors we've published previously, and then zero them out when we publish new errors. 

Ah that sounds good. Does this mean that we'll only need to re-query on saved instead of opens?

gbleaney: @grievejia I think we would want to query on open so that we tell the user right away if they open a file and it already has errors.

gbleaney:
@m0mosenpai there are two tests failing:
https://github.com/facebook/pyre-check/pull/433/checks?check_run_id=3111924028
https://github.com/facebook/pyre-check/pull/433/checks?check_run_id=3111924020

I believe you should be able to tweak both of the tests to do the import from one level higher (like you're doing locally). That should unblock landing this PR

client/commands/v2/pysa_server.py
grievejia: Oh I see what you mean. 

What I had in mind was to separate the querying logic from displaying logic, i.e. we do a query as soon as the background server establishes, and updates the query whenever a file is saved. Whether the query result gets shown to the users or not, on the other hand, gets controlled by the open request (or not at all if you want to show errors in unopened files as well).

But I think it would also work if we always query on open. Functionally they should be the same.

m0mosenpai:
> @m0mosenpai there are two tests failing:
> https://github.com/facebook/pyre-check/pull/433/checks?check_run_id=3111924028
> https://github.com/facebook/pyre-check/pull/433/checks?check_run_id=3111924020
> 
> I believe you should be able to tweak both of the tests to do the import from one level higher (like you're doing locally). That should unblock landing this PR

@gbleaney 
I think i'm confused how to fix the tests. I see that `pyre_test.py` and the imports in `__init__.py` in `tests` folder are failing. I was under the assumption that the tests would fail since the top-level hack i did was only done locally (running `pyre` as `python -m pyre-check.client.pyre` ). I'm not sure how to replicate the behavior for the tests.

gbleaney:
@m0mosenpai This is the command that the first test is running:
https://github.com/facebook/pyre-check/runs/3289612419?check_suite_focus=true#step:5:1

If you look into the script that runs that test, you can see that it just `cd`s into the root of the repo, and runs the tests in the `client` folder:
https://github.com/facebook/pyre-check/blob/main/scripts/run-python-tests.sh#L10-L13

You should be able to just tweak that test to `cd` up one directory higher, and run the tests in `pyre-check/client` rather than just `client`

This is the command that the second test is running:
https://github.com/facebook/pyre-check/runs/3289612418?check_suite_focus=true#step:8:4

You can see that is just runs a second script:
https://github.com/facebook/pyre-check/blob/main/documentation/deliberately_vulnerable_flask_app/run_integration_tests.sh#L7-L9

That test invokes the Pyre CLI from inside the `pyre-check` folder:
https://github.com/facebook/pyre-check/blob/e40e8c2d2de0b81a1ae244924c4f6a5647cce463/tools/pysa_integration_tests/utils.py#L156-L161
https://github.com/facebook/pyre-check/blob/e40e8c2d2de0b81a1ae244924c4f6a5647cce463/tools/pysa_integration_tests/utils.py#L177

You'll just need to change that to invoke from one dir higher, just like you did locally

gbleaney:
@m0mosenpai just checking, you're still planning to address this comment thread, right?
https://github.com/facebook/pyre-check/pull/433/files#r673484711

m0mosenpai:
@gbleaney  Yeah, I am

client/commands/v2/pysa_server.py
```diff
@@ -102,15 +169,22 @@ async def process_open_request(
         self, parameters: lsp.DidOpenTextDocumentParameters
     ) -> None:
         document_path = parameters.text_document.document_uri().to_file_path()
+        self.file_tracker[str(document_path)] = True
```
gbleaney, line 172: Any time you're only storing `True` in a dict, that's a good sign that you probably don't actually need a `dict`.  Instead, if you just want to see if something is in a collection or not (and maintain uniqueness), you could be using a `set`.

client/commands/v2/pysa_server.py
```diff
@@ -102,15 +169,22 @@ async def process_open_request(
...
+        # await self.log_and_show_message_to_client(f"Tracking Files: {self.file_tracker}")
```
gbleaney, line 173: We shouldn't be leaving any commented out logging in files we want to commit (apply throughout the PR)

client/commands/v2/pysa_server.py
gbleaney, line 172: Instead of tracking what files are open, why don't we track what files we've previously published diagnostics for? That would give us a much more exact picture of what files need their previous diagnostics overwritten. I imagine you could just update `file_tracker` in `update_errors` after this line:
```
diagnostics = self.invalid_models_to_diagnostics(model_errors)
```

client/commands/v2/pysa_server.py
```diff
@@ -52,13 +58,68 @@ def __init__(
         pyre_arguments: start.Arguments,
         binary_location: str,
         server_identifier: str,
+        file_tracker: dict,
```
gbleaney, line 61: Given that you don't actually ever pass anything but an empty dict into this constructor, there isn't any value in accepting `file_tracker` as a parameter. Instead, you can just create an empty set (see my comment below) and assign it to `self.file_tracker` all within the constructor.
```suggestion
```

gbleaney (changes requested):
Thanks for the update. A few more tweaks and hopefully we can land this!

client/commands/v2/pysa_server.py
m0mosenpai: Would that be possible? Since the errors are being calculated for all files at once, how will we know which files to track here?

client/commands/v2/pysa_server.py
```diff
@@ -59,6 +65,57 @@ def __init__(
         self.pyre_arguments = pyre_arguments
         self.binary_location = binary_location
         self.server_identifier = server_identifier
+        self.pyre_connection = api_connection.PyreConnection(
+            Path(self.pyre_arguments.global_root)
+        )
+        self.file_tracker = set()
```
gbleaney, line 71:
```suggestion
        self.file_tracker: Set[Path] = set()
```

client/commands/v2/pysa_server.py
```diff
@@ -59,6 +65,61 @@ def __init__(
...
+        # Publishing empty diagnostics to clear errors in VSCode and reset self.file_tracker
+        for document_path in self.file_tracker:
+            await _publish_diagnostics(self.output_channel, document_path, [])
+        self.file_tracker = set()
+
+        try:
+            model_errors = query.get_invalid_taint_models(self.pyre_connection)
+            diagnostics = self.invalid_models_to_diagnostics(model_errors)
+            # Keep track of files we publish diagnostics for
+            for path in diagnostics.keys():
```
gbleaney, line 115: No need to loop. `set.update` lets you provide a iterable to add into the set
```suggestion
            self.file_tracker.update(diagnostics.keys())
```

client/commands/v2/pysa_server.py
```diff
@@ -59,6 +65,61 @@ def __init__(
...
+                )
+            else:
+                result[Path(model.path)].append(self.invalid_model_to_diagnostic(model))
+        return result
+
+    async def update_errors(self) -> None:
+        # Publishing empty diagnostics to clear errors in VSCode and reset self.file_tracker
+        for document_path in self.file_tracker:
+            await _publish_diagnostics(self.output_channel, document_path, [])
+        self.file_tracker = set()
```
gbleaney, line 109: Very minor optimization, but there's no need to create a new set when we can just toss out everything in the old one
```suggestion
        self.file_tracker.clear()
```

gbleaney (changes requested):
Looks great! Two very small tweaks and then I think it's good from my perspective!

client/commands/v2/pysa_server.py
```diff
@@ -11,7 +11,12 @@
 
 import asyncio
 import logging
+from collections import defaultdict
+from pathlib import Path
+from typing import List, Sequence, Dict, Set
 
+from ....api import query, connection as api_connection
+from ....api.connection import PyreQueryError
```
grievejia, line 19: I would avoid relying on things in the `api` package -- those are considered "external" APIs that are currently going through multiple redundant levels of indirections (e.g. shelling out to `pyre` executable, which may or may not match the trunk version). For v2 command, there's no need to shell out to anything: we could query the background Pyre server by invoking command-internal logic directly. 

Querying the Pyre server in pure Python can be done with `commands.v2.query.query_server`:
```
from . import query

try:
  socket_path = Path(self.pyre_arguments.base_arguments.log_path)  # You probably need a rebase to get `base_arguments` to appear
  response_json = query.query_server(socket_path, "validate_taint_models()").payload
except query.InvalidQueryResponse:
  ...
```
Some followup logic may be needed to extract useful info from the `response_json` dict.

PR #440, opened 2021-06-30
Adds publishable github action for pysa

abishekvashok:
Adds publishable github action that others can include into their github
actions that enables users to run pysa via github actions.

The action is a composite github action that installs pyre-check,
runs pysa, and expose the results in SARIF format by creating sarif.json
at the repository root using fb-sapp.

Adds instructions to run the action in
documentation/pysa_github_action/README.md

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/13

See it in action: https://github.com/abishekvashok/zulip/actions/runs/1149735960


abishekvashok:
There were listing issues, fixed them :D

abishekvashok:
Here's a PR I made from my Zulip fork to itself that demonstrates the working of the changes in the PR:
https://github.com/abishekvashok/zulip/pull/2

Scroll down to checks sections of the PR and see Code Analysis Pysa there (at place where SARIF results show up)
Screenshot 2021-07-05 at 11 04 17 PM

Clicking on which should lead you to the issues list. Github is robust enough to show even difference in issues!
Screenshot 2021-07-05 at 11 04 43 PM

action.yml
```diff
@@ -0,0 +1,18 @@
...
+    required: false
+    default: "stable"
+runs:
+  using: "composite"
+  steps:
+    - run: if [ ${{inputs.build-type}} = "nighty" ]; then pip3 install pyre-check-nightly; else pip3 install pyre-check; fi
+      shell: bash
+    - run: if [ ! -f .pyre_configuration ]; then printf ".\n" | pyre init; fi
+      shell: bash
+    - run: pyre -n analyze --no-verify | python3 ${{github.action_path}}/scripts/convert_pysa_results_to_sarif.py > sarif.json
```
r0rshark, line 16: Internally @0xedward is working on supporting sarif output directly inside SAPP. This allow us to use filters and not simply output all the issues. @abishekvashok we can put this diff on hold until we have this landed and then we can use the new SAPP feature to output SARIF results and then make use of it here. This will probably mean that instead of calling "python3 ${{github.action_path}}/scripts/convert_pysa_results_to_sarif.py " we'll call sapp with some parameter. What do you think?

r0rshark (changes requested):
Nice I like the documentation! Let's wait for the SAPP feature to be implemented :)

action.yml
abishekvashok: Makes sense to me!

action.yml
```diff
@@ -0,0 +1,18 @@
...
+    description: Pysa build to use. Possible values are stable and nightly. Default is stable.
+    required: false
+    default: "stable"
+runs:
+  using: "composite"
+  steps:
+    - run: if [ ${{inputs.build-type}} = "nighty" ]; then pip3 install pyre-check-nightly; else pip3 install pyre-check; fi
+      shell: bash
+    - run: if [ ! -f .pyre_configuration ]; then printf ".\n" | pyre init; fi
+      shell: bash
```
r0rshark, line 15: We can try to run `pyre infer` before launching pysa. With this pyre will try to infer more type information and Pysa may be able to find more issues https://pyre-check.org/docs/pysa-coverage/#pyre-infer

abishekvashok: Yeah, we can try that!

abishekvashok: Did it!

abishekvashok:
@r0rshark the changes have been made to run pyre infer as well, and it seems the suggestion was an awesome one indeed! The linked pull request has these changes and we detecting large no: issues in Zulip.

Thanks for that! Let's wait till sapp gets SARIF support now :)

abishekvashok:
@0xedward I see the sarif changes landed on sapp 🥳❤️
Any idea as to when a version dump would be done? Or are there going to be more changes?

0xedward:
> Any idea as to when a version dump would be done? Or are there going to be more changes?

@abishekvashok thanks for staying on top of this! :) [`fb-sapp` 0.5.0](https://pypi.org/project/fb-sapp/0.5.0/) has SARIF support  

The SAPP docs will be updated soon (after a few more changes roll out), but here's the workflow you can use to get SARIF output right now:
1. `pip install fb-sapp`
1. `sapp analyze /path/to/pysa/taint-output.json`
2. `sapp update warning-codes /path/to/pysa/taint-metadata.json`
3. `sapp filter issues 1 /path/to/pysa/filters --output-format sarif`
4. If you want to save the SARIF output to a file, `sapp filter issues 1 /path/to/pysa/filters --output-format sarif > filtered-pysa-output.sarif`

You probably already have your environment all set up, but if you haven't installed `fb-sapp` before, here's some [reference docs for common issues](https://pyre-check.org/docs/pysa-quickstart/#common-issues) folks run into during the process

abishekvashok:
@0xedward thanks for the reply. Made the required changes to use sapp. 

CC: @r0rshark @gbleaney 

Caution: I read somewhere recently that only a single publishable action can be contained in a repository. If we require to add a publishable action to pyre, then this would prove a blocker (migration would destroy existing actions). Also, since we are adding this action to a repository of a project, we can't publish this to GitHub action marketplace. If we want to do that, the actions.yml file needs to be in a repository of its own (ref. https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace#about-publishing-actions ). Also either way, consider adding a branch for actions that freeze `pyre-check` like v1, v2 etc. (or can we make use of stable?) (this can be skipped, but we can end-up breaking stuff, eg: this broke when `pyre infer -r` was merged into `pyre infer` making the former a error)

0xedward:
> @0xedward thanks for the reply. Made the required changes to use sapp. Using no filters however leads the issues to 0, do you know how to run without filters @0xedward ?

I don't think SAPP currently has a way to get a list of issues for a particular run without applying filters. I'm not sure if adding that functionality would be useful for us since adding that feature would just be changing the json output from what Pysa, but the data in the output would be the same. Let me know if this is a blocker though and I can add this.

Regarding SAPP outputting 0 issues after applying filters, [the output from the GitHub action run shows this](https://github.com/abishekvashok/zulip/pull/2/checks?check_run_id=3360795309):
```
Usage: sapp filter issues [OPTIONS] RUN_ID INPUT_FILTER_PATH
Try 'sapp filter issues --help' for help.

Error: Missing argument 'INPUT_FILTER_PATH'.
Error: Process completed with exit code 2.
```

Exit code 2 here means that the path doesn't exist. I see that you ran `mkdir ./.pysa/` earlier in the workflow. My guess here is that this issue is related to using a composite Github Action

> Caution: I read somewhere recently that only a single publishable action can be contained in a repository. If we require to add a publishable action to pyre, then this would prove a blocker (migration would destroy existing actions). Also, since we are adding this action to a repository of a project, we can't publish this to GitHub action marketplace. If we want to do that, the actions.yml file needs to be in a repository of its own (ref. https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace#about-publishing-actions ). Also either way, consider adding a branch for actions that freeze pyre-check like v1, v2 etc. (or can we make use of stable?) (this can be skipped, but we can end-up breaking stuff, eg: this broke when pyre infer -r was merged into pyre infer making the former a error)

Thanks for bringing this up! I'm not sure what others think about creating a branch for the GitHub action, since we will need to include the Github Action `yml` file in the root of the repo. Like you mentioned, I'm guessing the Pyre folks would like to publish a GitHub action as well. I'll check with them to see what they prefer.

Another option for us to consider is to move create a single GitHub action under SAPP and make the action configurable for any static analysis tools we use with SAPP.

action.yml
```diff
@@ -0,0 +1,30 @@
...
+    - run: pip install fb-sapp
+      shell: bash
+    - run: mkdir ./.pysa/
+      shell: bash
+    - run: pyre analyze --no-verify --save-results-to ./
+      shell: bash
+    - run: sapp analyze ./taint-output.json
+      shell: bash
+    - run: sapp filter issues 1 ./.pysa/ --output-format sarif > sarif.json
```
0xedward, line 29: If the `./.pysa` directory here exists, I think the `./.pysa/` would be empty, resulting in no filters applied to the run.

The pysa filters are included with `pyre-check` package. It looks like in this case the filters are in the following directory: `/opt/hostedtoolcache/Python/3.9.6/x64/lib/pyre_check/pysa_filters`.

I recommend copying the setup instructions for Ubuntu from the [quickstart guide](https://pyre-check.org/docs/pysa-quickstart/) for this Github Action (e.g. use a virtual environment)

documentation/pysa_github_action/README.md
```diff
@@ -0,0 +1,69 @@
+# Pysa Github Action
+
+Run pysa on the repository in Github Actions and create a SARIF File.
+
+## Dependencies
+
+The action requires a python setup with the python version > 3.5
```
0xedward, line 7: Is there a way to specify we want a python version > 3.7? SAPP won't run on python versions earlier than 3.7, because of some type annotations we use

action.yml
```diff
@@ -0,0 +1,30 @@
...
+runs:
+  using: "composite"
+  steps:
+    - run: if [ ${{inputs.build-type}} = "nighty" ]; then pip install pyre-check-nightly; else pip install pyre-check; fi
+      shell: bash
+    - run: if [ ! -f .pyre_configuration ]; then printf ".\n" | pyre init; fi
+      shell: bash
+    - run: pyre infer
+      shell: bash
+    - run: pyre infer -i --annotate-from-existing-stubs
```
0xedward, line 19: It's a good idea to use `pyre infer`, but since we are using `infer` in-place, we might want to include in docs somewhere that the output from SARIF will be different from their code (e.g. line numbers in traces being off, type annotations that the developers didn't add themselves, etc)

abishekvashok:
> @abishekvashok I don't think SAPP currently has a way to get a list of issues for a particular run without applying filters. I'm not sure if adding that functionality would be useful for us since adding that feature would just be changing the json output from what Pysa, but the data in the output would be the same.

Yeah makes sense. I will try adding an additional flag for SAPP filters to use with pysa. But I am confused as to if I do it, everyone who makes use of pysa should use at least a filter... But we can get around it by making it optional and adding a filter that passes everything through or something?

>  I'll check with them to see what they prefer and get back to you on this :)

@0xedward that will be a huge help! Thank you.

> Another option for us to consider is to move create a single GitHub action under SAPP and make the action configurable for any static analysis tools we use with SAPP.

Yeah that would be awesome. In which case we could edit the issue, https://github.com/MLH-Fellowship/pyre-check/issues/13 towards the same goal if we get a suitable response from the pyre-check team :)

action.yml
```diff
@@ -0,0 +1,30 @@
...
+    default: "stable"
+
+runs:
+  using: "composite"
+  steps:
+    - run: if [ ${{inputs.build-type}} = "nighty" ]; then pip install pyre-check-nightly; else pip install pyre-check; fi
+      shell: bash
+    - run: if [ ! -f .pyre_configuration ]; then printf ".\n" | pyre init; fi
+      shell: bash
+    - run: pyre infer
```
0xedward, line 17: Before running `pyre infer`, we should also make sure we install the repo's dependencies the action is running on with `pip install -r requirements.txt`

[Step 2 in Initial configuration in the quickstart guide](https://pyre-check.org/docs/pysa-quickstart/#initial-configuration) provides context for why we need to do this

0xedward (changes requested):

action.yml
abishekvashok: Yeah I made a temporary directory as a last resort to see if giving an empty directory allows us to pass through. But it didn't and returned 0 issues. I will check the filters we have with pyre_check :)

documentation/pysa_github_action/README.md
abishekvashok: The composite action makes use of the existing version of python the action uses. So the thing we can do is, make a comment. We can maybe, migrate this action to a node js action if we require total control.

action.yml
abishekvashok: Yep makes sense :)

action.yml
abishekvashok: Okay.

0xedward:
> Yeah makes sense. I will try adding an additional flag for SAPP filters to use with pysa. But I am confused as to if I do it, everyone who makes use of pysa should use at least a filter... But we can get around it by making it optional and adding a filter that passes everything through or something?

@abishekvashok if we did want to implement it, we would just query for all the issues corresponding to a run id the user provides rather than creating a filter allows all issues to pass through, but I don't think this is worth doing, because I don't see a use case for this right now. I think we can add a feature like this if for some reason a user wants to convert the json format pysa outputs to the json format sapp outputs, but I'm not sure who would want to do that right now.

The issue here doesn't seem related to SAPP. I think it's because the directory `./.pysa` doesn't exist. Here's the exact line the issue here is stopping on - https://github.com/facebook/sapp/blob/main/sapp/cli_lib.py#L302

documentation/pysa_github_action/README.md
0xedward: Can we use the following here?
```
    strategy:
      matrix:
        python-version: [3.6, 3.7, 3.8, 3.9]
```

https://docs.github.com/en/actions/guides/building-and-testing-python#starting-with-the-python-workflow-template

abishekvashok:
> The issue here doesn't seem related to SAPP not applying filters to runs. I think it's because the directory ./.pysa doesn't exist. Here's the exact line the issue here is stopping on - https://github.com/facebook/sapp/blob/main/sapp/cli_lib.py#L302

I did try the other way, the diff currently was updated again (I often rewrite git history to prevent 100s of commits from appearing) to try the current approach.

The goal of the action is to have the issues pysa reports in a sarif format so that users can directly see the errors in the diff as GitHub supports sarif. https://github.com/MLH-Fellowship/pyre-check/issues/13

I will think about it a bit, and confirm with @r0rshark when he comes back next week, if we need sapp in the first place. Before sapp was in place here, we converted the json results pysa gave out to sarif format using a script similar to how pyre output was in sarif (but now migrated via my latest pr), thanks for your time and quick response which I appreciate. I will think and get back :)

arthaud:
Could you please rebase this on master and fix the lint error?

0xedward:
> Could you please rebase this on master and fix the lint error?

It's alright! We can rebase this internally before landing @abishekvashok fixed the lint error in https://github.com/facebook/pyre-check/commit/91e9a2afcd7db13099b22a91d3c4cb6560a26a08 :)

0xedward:
> @0xedward I did try the other way, the diff currently was updated again (I often rewrite git history to prevent 100s of commits from appearing) to try the current approach.

@abishekvashok oh, sorry I just looked at the latest commit on this PR and the latest Github Actions run on https://github.com/abishekvashok/zulip/pull/2/checks?check_run_id=3360795309. Would you be able to link me to the yaml file and the respective run where you are getting this result, so I can help debug?

Another way to debug this is to manually run the same list of commands you have in the yaml file to see what's happening at each step. (e.g. git clone zulip, set up venv, install pyre-check and fb-sapp, run pysa, ingest results with sapp, etc)

> The goal of the action is to have the issues pysa reports in a sarif format so that users can directly see the errors in the diff as GitHub supports sarif. MLH-Fellowship#13
> I will think about it a bit, and confirm with @r0rshark when he comes back next week, if we need sapp in the first place. Before sapp was in place here, we converted the json results pysa gave out to sarif format using a script similar to how pyre output was in sarif (but now migrated via my latest pr), thanks for your time and quick response which I appreciate. I will think and get back :)

We had chats internally about how Pysa sometimes can provide a lot of false positives out of the box, so running the results through SAPP to filter a lot of false positives from the list of issues we eventually upload in SARIF would reduce alert fatigue.

Like you mentioned, https://github.com/MLH-Fellowship/pyre-check/issues/13 only mentions creating an action for Pysa. I'm not sure if the intention here was to create two Github Actions (one for pysa without sapp, and one with pysa with sapp), so it makes sense to check with @r0rshark about this :)

action.yml
```diff
@@ -0,0 +1,30 @@
...
+      shell: bash
+    - run: pyre infer -i --annotate-from-existing-stubs
+      shell: bash
+    - run: pip install fb-sapp
+      shell: bash
+    - run: mkdir ./.pysa/
+      shell: bash
+    - run: pyre analyze --no-verify --save-results-to ./
+      shell: bash
+    - run: sapp analyze ./taint-output.json
```
0xedward, line 27: We also need to run `sapp update warning-codes ./taint-metadata.json` here before running `sapp filter issues`. 

The `update warning-codes` command adds the list of warning codes present in the Pysa run into SAPP db to reference when creating the rules property in the SARIF output

0xedward (changes requested):
Not sure why this comment didn't go out with my previous review :(

documentation/pysa_github_action/README.md
abishekvashok: Unfortunately that is for a GitHub action -> not something that we can use here (because composite actions don't support uses statement yet). However, I will add this as a comment.

abishekvashok:
@0xedward @arthaud made the requested changes and rebased on top of master.

abishekvashok:
@0xedward @arthaud made the requested changes (also rebased with main, to avoid you more trouble :) ) 

Sorry for the back and forth.

You may view the action here: https://github.com/abishekvashok/zulip/actions/runs/1149735960
The results of the PR: https://github.com/abishekvashok/zulip/pull/2/checks?check_run_id=3379365809

PS: I haven't talked with Lorenzo yet (just to make sure this is what he intended :) ), so if you can wait till early next week, it would be awesome. Also keep in mind that this change can lead to no further action being added to this repository (changing actions can cause other people's builds to fail :))

0xedward:
Closing this PR since we have [released a GitHub Action for Pysa](https://github.com/marketplace/actions/pysa-action)

Thanks for your early work on this, @abishekvashok :)

abishekvashok:
Yayyy

PR #442, opened 2021-07-06
Adds Via feature to flask.Markup.escape

abishekvashok:
Adds via escape_html feature to flask.Markup.escape

The other possible sink for flask.Markup was already implemented as both
flask and Jinja2 share the same library.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/15

Before the change: 
Screenshot 2021-07-06 at 4 34 18 PM

After the change:

PS: The `after the change` screenshot was updated after the latest changes.


stubs/third_party_taint/xss_sinks.pysa
```diff
@@ -6,6 +6,9 @@ def markupsafe.Markup.__new__(cls, base: TaintSink[XSS], encoding, errors) : ...
 @SkipObscure
 def markupsafe._native.escape(s: TaintInTaintOut[Via[escape_html]]): ...
 
+# Markup library used by Flask
+def markupsafe.Markup.escape(s: TaintInTaintOut[Via[escape_html]]): ...
```
gbleaney, line 10: Two tweaks we should make here:
1) Adding `@SkipObscure`. This is something we just released, and is documented here: https://pyre-check.org/docs/pysa-advanced#obscure-models . We want `SkipObscure` as a temporary workaround for this issue that's preventing us from getting `always-via:escape_html` (you'll notice in your test we're only getting `via:escape_html`)
2) Adding `LocalReturn`. This is so we actually propagate taint from `s` to the return value, despite the `SkipObscure` we're adding. Documented here: https://pyre-check.org/docs/pysa-basics#taint-propagation

```suggestion
@SkipObscure
def markupsafe.Markup.escape(s: TaintInTaintOut[LocalReturn, Via[escape_html]]): ...
```

gbleaney (changes requested):
Thanks for the PR! And I really appreciate that you included evidence of your testing :)

abishekvashok:
@gbleaney thanks for the suggestions, the changes have been made!

Screenshot 2021-07-07 at 10 09 22 AM

gbleaney:
Thanks @abishekvashok! One final thing: can you update your test plan? I believe it should show an `always-via:escape_html` now, and I'd like to confirm that's the case

abishekvashok:
>  I believe it should show an always-via:escape_html now, and I'd like to confirm that's the case

@gbleaney Hmm, I am a bit clueless now. I believe I got it in the screenshot posted above. I'll repost again,



But I don't know -- sorry -- were you suggesting I run a specific test or something?

abishekvashok:
Upss sorry @gbleaney, I should have used the same test plan! Here's the output anyway, I hope this is what you seek :)
Screenshot 2021-07-07 at 7 53 35 PM

gbleaney:
Yup, looks perfect now :)

Issue #444, opened 2021-07-07
Pysa Tutorial missing fb-sapp


jckling:
In exercise4 I'm told to use `sapp`  but it isn't installed before. So I need to `pip3 install fb-sapp` before I continue the exercise.

The [README.md](https://github.com/facebook/pyre-check/tree/master/documentation/pysa_tutorial/exercise4#static-analysis-post-processor-sapp) file in exercise4 said:
> The Static Analysis Post Processor (SAPP) ships in a separate package from Pysa, which you should have already installed during the initial setup of this tutorial. 

Is it missing? SAPP should be installed in **Setup Instructions**.

gbleaney:
Thanks for the report! This problem occurred because we removed `pyre-check`s dependency on `fb-sapp` a while ago, but forgot to update the tutorial. I've put up #446  to fix this.

PR #446, opened 2021-07-07
Remove redundant pip install and add install for sapp

gbleaney:
Issue #444 pointed out that SAPP is missing from the venv when doing the tutorial. This is because we removed `pyre-check`s dependency on `fb-sapp`, so that users of Pyre wouldn't have to bring in a whole web server when all they wanted was a type checker. This change updates our `Dockerfile` and `README.md` to now explicitly install `fb-sapp`.

Additionally, we've fixed the dependencies on the `fb-sapp` package, and we no longer need to explicitly install dependencies like `flask` or `click`. For that reason, I've also removed the now redundant extra `pip install` lines.

Test Plan:
Removed my old venv:
```
rm -rf tutorial/
```
Set up a new one with just `pyre-check` and `fb-sapp`:
```
pip3 install pyre-check fb-sapp
```
Ran exercise 4, which requires both:
```
pyre analyze --no-verify --save-results-to .
sapp analyze taint-output.json
sapp explore
```
Was able to interactively explore models in SAPP

### DIFF
--- a/documentation/pysa_tutorial/Dockerfile
+++ b/documentation/pysa_tutorial/Dockerfile
@@ -14,5 +14,4 @@ WORKDIR /pyre-check
 COPY . .
 
 RUN pip3 install psutil
-RUN pip3 install pyre-check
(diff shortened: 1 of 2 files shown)


abishekvashok (approved):
LGTM

PR #448, opened 2021-07-12
Adds sinks for pymysql

abishekvashok:
Adds SQL sinks for python MySQL library/module regarding functions
`execute` and `executemany`.

Fixes https://github.com/MLH-Fellowship/pyre-check/issues/20


abishekvashok:
PyMysql stubs are in type shed, here are the screenshots
Before this change:
pymysql before

After this change:
pymysql after 1
pymysql after 2

abishekvashok:
The stubs for PyMysql are in type shed.

Before this change:
pymysql before

After this change
pymysql after 1
pymysql after 2

r0rshark:
Great! Did you check the API to see if there are more functions which may lead to a SQL injection? for example doe this model work for the [DictCursor](https://pymysql.readthedocs.io/en/latest/modules/cursors.html#pymysql.cursors.DictCursor) object as well

abishekvashok:
> Great! Did you check the API to see if there are more functions which may lead to a SQL injection? for example doe this model work for the DictCursor object as well

Yes I did, there are various Cursors all inheriting the execute and executemany functions without a change. Since we have stubs for the parent class, I guess this should work.

r0rshark:
Yes correct. Just wanted to make sure they are inherited from a base class :)

PR #449, opened 2021-07-12
Adds sinks for pymssql

abishekvashok:
pymssql is a popular sql python library/module. Adds pysmssql SQL sinks
for the functions `execute` and `executemany`.

Fixes https://github.com/MLH-Fellowship/pyre-check/issues/21


abishekvashok:
The stubs are in typeshed

Before this change:
pymssql - before

After this change:
pymssql - after 1
pymssql-after

r0rshark:
Great work! Can you have a look at the [_mssql](https://www.pymssql.org/ref/_mssql.html#_mssql.MSSQLConnection.execute_query) module which seems to be shipped with pymssql as well?

abishekvashok:
> Can you have a look at the _mssql module which seems to be shipped with pymssql as well?

Will take a look and get back!

abishekvashok:
Hey @r0rshark so I did take a look and tried to get the models in. It seems the stubs of pymyssql, https://github.com/python/typeshed/blob/master/stubs/pymssql/%40python2/pymssql.pyi , don't contain stubs for the `_mssql` module. Can we include the stubs as we do for `lxml` and others? Just wanted to affirm that it would be the suitable approach :)

r0rshark:
Bummer this is not in typeshed :( . I would avoid adding additional custom stubs. I do not see much usage of the `_mssql ` module directly (most of the time people seem to use pymssql) so it may not be a big deal if we do not have these models. 
However if you want to try the best approach is to create a pull request to [python/typeshed](https://github.com/python/typeshed) with the _mssql typing information following the instruction in https://github.com/python/typeshed/blob/master/CONTRIBUTING.md .I think a good starting point is to autogenerate the pyi using mypy stubgen (https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#using-stubgen)

abishekvashok:
> However if you want to try the best approach is to create a pull request to python/typeshed with the _mssql typing information following the instruction in https://github.com/python/typeshed/blob/master/CONTRIBUTING.md .I think a good starting point is to autogenerate the pyi using mypy stubgen (https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#using-stubgen)

I will make a try, but as you said, we can go without `_mssql` till then :)

abishekvashok:
@r0rshark I tried to run _mssql without stubs, but it isn't getting included in the environment as mentioned in the discord message.
Screenshot 2021-07-14 at 6 46 34 PM

PR #450, opened 2021-07-12
Adds missing models and stubs for lxml

abishekvashok:
Adds missing models and stubs for lxml.

Modifies existing stubs of lxml function `lxml.etree.fromstring` to
include the optional arguments `parser` and `base_url`.

Adds stubs for the function `lxml.etree.fromstringlist`

Adds Pysa models for the function `lxml.etree.fromstringlist` and
declare its input string list argument as an `XMLParser` sink.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/23


abishekvashok:
lxml stubs have been modified to further reflect the stubs they maintain: https://github.com/lxml/lxml-stubs/blob/master/lxml-stubs/etree.pyi. Those stubs are not available in typeshed due to some issues. However, the current stubs can be removed when they get integrated into typeshed so we won't have duplicates. For now, I added additional definitions Pysa. This helps us catch issues when we specify the optional parser parameter which otherwise Pysa won't catch.

Before the changes:
lxml-b-1
lxml-b-2
lxml-b3

After the changes:
lxml-a-1
lxml-a-2
lxml-a-3

stubs/third_party_taint/lxml_sinks.pysa
```diff
@@ -1,2 +1,4 @@
-def lxml.etree.fromstring(input: TaintSink[XMLParser]): ...
+def lxml.etree.fromstring(input: TaintSink[XMLParser], parser, *, base_url): ...
```
gbleaney, line 1: These extra parameters are actually not required by Pysa, and they make the model harder to maintain (more things to keep in sync with the stub), so we prefer to not include them.

(Apply throughout the diff)

```suggestion
def lxml.etree.fromstring(input: TaintSink[XMLParser]): ...
```

gbleaney (changes requested):

abishekvashok:
@gbleaney Done :D

PR #452, opened 2021-07-13
Right-click to generate model

gracewgao:
Refer to [[Fall 2021] Step 3: VSCode Plugin - right click to generate model](https://github.com/MLH-Fellowship/pyre-check/issues/7). 

<details>
<summary>Screenshots</summary>

Right-click menu screenshot
Screenshot of popup message

</details>


gracewgao:
Also, this is what I'm currently stuck on:
image

@m0mosenpai @gbleaney  Did I miss something with the configuration?

api/query.py
```diff
@@ -253,3 +267,31 @@ def get_invalid_taint_models(
                 )
             )
     return errors
+
+
+def types(pyre_connection: PyreConnection, modules: Iterable[str]) -> List[Types]:
```
gbleaney, line 272: Please add tests for this function. Tests are here:
https://github.com/facebook/pyre-check/blob/11ab586bd77f97c5778d89a9338f178203569ee1/api/tests/query_test.py

gbleaney, line 272: This is really a collection of paths, not modules, right?

```suggestion
def types(pyre_connection: PyreConnection, paths: Iterable[str]) -> List[Types]:
```

client/commands/v2/language_server_protocol.py
```diff
@@ -424,3 +424,18 @@ def from_json_rpc_parameters(
...
             parameters, target=WorkspaceDidChangeConfigurationParameters
         )
+
+@dataclasses_json.dataclass_json(
+    letter_case=dataclasses_json.LetterCase.CAMEL,
+    undefined=dataclasses_json.Undefined.EXCLUDE,
+)
+@dataclasses.dataclass(frozen=True)
+class DidCopyModelParameters:
+    path: Path
```
gbleaney, line 434: This will need to be a str, and then converted to `Path` after, for serialization purposes
```suggestion
    path: str
```

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,77 @@ def __init__(
         self.binary_location = binary_location
         self.server_identifier = server_identifier
 
+    def invalid_model_to_diagnostic(
```
gbleaney, line 68: `invalid_model_to_diagnostic`, `invalid_models_to_diagnostics`, and `update_errors` are present because you're based off of Sarthak's branch, right? You didn't change these?

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,77 @@ def __init__(
...
+            model_errors = query.get_invalid_taint_models(pyre_connection)
+            diagnostics = self.invalid_models_to_diagnostics(model_errors)
+            await self.show_model_errors_to_client(diagnostics)
+        except PyreQueryError as e:
+            await self.log_and_show_message_to_client(
+                f"Error querying Pyre: {e}", lsp.MessageType.WARNING
+            )
+
+    async def copy_model(self, document_path: Path, position: lsp.Position) -> None:
+        pyre_connection = api_connection.PyreConnection(
```
gbleaney, line 119: It probably makes sense to store this connection as an attribute of the server, rather than re-making it every time.

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,77 @@ def __init__(
...
+            Path(self.pyre_arguments.global_root)
+        )
+        try:
+            types = query.types(pyre_connection, [document_path])
+            for type in types[0].types:
+                stop = lsp.Position(
+                    line=type.location["stop"]["line"],
+                    character=type.location["stop"]["column"],
+                )
+                if stop.line > position.lline or (
```
gbleaney, line 129: Typo?

```suggestion
                if stop.line > position.line or (
```

client/commands/v2/pysa_server.py
```diff
@@ -60,6 +65,77 @@ def __init__(
...
+            )
+
+    async def copy_model(self, document_path: Path, position: lsp.Position) -> None:
+        pyre_connection = api_connection.PyreConnection(
+            Path(self.pyre_arguments.global_root)
+        )
+        try:
+            types = query.types(pyre_connection, [document_path])
+            for type in types[0].types:
+                stop = lsp.Position(
```
gbleaney, line 125: Why use `lsp.Position` here? It doesn't seem to gain un anything. If you switch to `query.Position`, and implement comparators on that type, you can make the logic much cleaner.

api/query.py
```diff
@@ -47,6 +47,20 @@ class Location(NamedTuple):
     stop: Position
 
 
+class Type(NamedTuple):
+    location: Dict[str, Position]
+    annotation: str
+
+    def extract_function_model(self) -> str:
+        results = re.findall("\(.*?\)", self.annotation)
```
gbleaney, line 55: The annotation will look something like this:
```
"typing.Callable(views.operate_on_twos)[[Named(request, django.http.request.HttpRequest)], django.http.response.HttpResponse]"
```

It's probably a good idea to be a bit more strict on the validation (maybe require matching the preceding `typing.Callable`?)

api/query.py
```diff
@@ -47,6 +47,20 @@ class Location(NamedTuple):
...
+        return results[0] if results else ""
```
gbleaney, line 56: I would suggest an exception, rather than an empty string. If someone called this method, they thought the represented type was a function. If that assumption isn't true, we shouldn't hide that fact

api/query.py
gbleaney: Also note that right now you're just returning `views.operate_on_twos` from that example, but you really need to be returning `def views.operate_on_twos(request): ...`

client/commands/v2/pysa_server.py
gbleaney: I think this logic is too simple. I think we need to check that `position` is between the `start` and `stop` positions of `type`.

tools/ide_plugins/pysa-vscode/src/main.ts
```diff
@@ -9,57 +9,93 @@
  * @file Entry point for Pysa's VSCode extension.
  */
 
-import * as vscode from 'vscode';
```
gbleaney, line 12: It looks like you did a complete reformat of this file. That's fine, but formatting changes make it very hard to review functional changes. Best practice is to separate formatting changes into a different commit/PR

tools/ide_plugins/pysa-vscode/src/main.ts
```diff
@@ -9,57 +9,93 @@
...
+
+async function copyModel(path: string) {
+  try {
+    var active = vscode.window.activeTextEditor.selection.active;
+    languageClient
+      .sendRequest("copyModel", {
+        path: path,
+        position: { line: active.line, character: active.character },
+      })
+      .then((response: any) => {
```
gbleaney, line 88: Can we do better than `any` on typing here?

gbleaney (changes requested):

abishekvashok:
> Also, this is what I'm currently stuck on:

Sorry for popping in, but will add my experience here, idk if it helps, but I usually see this error when I don't have a `.pyre_configuration` at the directory I am running from...

client/commands/v2/pysa_server.py
gracewgao: Yes, these were pulled from Sarthak's branch!

gracewgao:
@abishekvashok Thanks for the tip! I think the issue here for me was very similar and it was because the `taint.config` file was missing in my directory.

client/commands/v2/pysa_server.py
gracewgao: Sure, I'll add that! I think I had it this way originally since the results are in order (but if anything changes with the `types` query then you're right that this wouldn't work

gracewgao:
@gbleaney Thanks for helping me debug earlier! The whole flow is working now:

Right-click menu

Copied model

I think what's left now is:
- writing test for the `types` query function
- adding a global exception into the `pysa_server.py` file
- tidying up the comments you left above from the code review
- pulling in Sarthak's branch (or would you prefer to do this later?)

I'm aiming to get these done this week, but let me know if there's anything else I missed!

gracewgao:
It's been a while since I had a chance to review this, but I took a bit of time to go through and try to fix some things if this is still a relevant PR :)

api/query.py
```diff
@@ -40,13 +50,45 @@ class Position(NamedTuple):
     line: int
     column: int
 
+    def __eq__(self, other) -> bool:
+        return self.line == other.line and other.column == self.column
+
+    def __gt__(self, other) -> bool:
+        return self.line > other.line or (
+            other.line == self.line and other.column <= self.column
```
gbleaney, line 58: Consistency makes it sooo much easier for the reader to understand code. So I'd suggest:

1. Always keeping to operands in the comparison in the same order (eg. `self` on left, `other` on right)
2. Since this is a implementation of `>`, use `>` in the comparisons rather than `<=`
Apply to the `__lt__` too

```suggestion
            self.line == other.line  and self.column > other.column
```

api/query.py
```diff
@@ -40,13 +50,45 @@ class Position(NamedTuple):
...
     start: Position
     stop: Position
 
 
+class Type(NamedTuple):
+    location: Dict[str, Position]
+    annotation: str
+
+    def extract_function_model(self) -> str:
+        functions = re.findall(r"(?<=typing.Callable\().*?(?=\))", self.annotation)
```
gbleaney, line 78: I think we can do better than a regex here. `pyre query "defines(package.module)"` will return you structured data with function names and parameters. That's going to be much more robust compared to using regex to pull data from a type annotation

client/commands/v2/pysa_server.py
```diff
@@ -129,60 +219,102 @@ async def process_did_save_request(
...
             )
+        await self.update_errors(document_path)
+
+    async def process_copy_model_request(
+        self, parameters: lsp.DidCopyModelParameters
+    ) -> str:
+        document_path = parameters.path
+        if document_path is None:
+            raise json_rpc.InvalidRequestError(
+                f"Document path is not a file: {parameters.path}"
```
gbleaney, line 230: Nit: That's not what you're checking. You're just checking that `document_path` is not `None` or `""`. I'd suggest tweaking the error (or going further an checking that `document_path` does indeed point to a file, so this error message is true)

client/commands/v2/pysa_server.py
```diff
@@ -129,60 +219,102 @@ async def process_did_save_request(
...
+            )
+        return await self.copy_model(document_path, parameters.position)
 
     async def run(self) -> int:
         while True:
             async with _read_lsp_request(
                 self.input_channel, self.output_channel
             ) as request:
-
```
gbleaney, line 138: This whole block of changes is actually just an indentation change plus the addition of the `"pysa/copyModel"` branch, right?

client/commands/v2/pysa_server.py
```diff
@@ -11,7 +11,12 @@
 
 import asyncio
 import logging
+from os.path import relpath
+from pathlib import Path
+from typing import List, Optional, Sequence, Dict
 
+from ....api import query, connection as api_connection
+from ....api.connection import PyreQueryError
```
gbleaney, line 19: @grievejia had some feedback about using `api` on Sarthak's PR:
https://github.com/facebook/pyre-check/pull/433/files#r704607152

Could you take a look at his suggested alternative?

gbleaney (changes requested):
Thanks so much for coming back to finish this up! A few more suggestions and then I'm excited to land this!

api/query.py
gracewgao: Hmm, definitely agree that regex doesn't seem like the cleanest solution, but I'm not sure how we can determine what we want to extract from `pyre query "defines(package.module)` without using regex on at least the annotation returned from the `types` query.

client/commands/v2/pysa_server.py
gracewgao: Yep! Just my linter being overdramatic as usual 🙄

client/commands/v2/pysa_server.py
gracewgao: Hmm, it looks like the new logic for querying `types` is also in the `api` package - would you recommend refactoring them into `query.py` or a separate file in `client.commands.v2`? Lmk if you have any other suggestions for going about this!

api/query.py
gbleaney: Fair point. We could still avoid the second regex using `defines`, but I think you're right we can't avoid the first regex with the way things are currently written.

client/commands/v2/pysa_server.py
gbleaney: My first thought would be a separate file, but I don't feel to strongly about it.

gbleaney (changes requested):
Putting back in requested changes status pending the comments. Again, thanks for working on this!

arthaud:
Closing this since there hasn't been any update since 2021. Feel free to re-open or make a new PR!

PR #453, opened 2021-07-14
Mirror internal linter behaviour in open source

abishekvashok:
Mirror the Facebook internally used linter/formatter behaviour in the
open source repository.

Pins to black and usort versions used in the internal linter and runs
black, flake8, and usort in seperate runner instances.

Adds pyproject.toml with identical usort configuration as used
internally.

Modifies flake8 configuration to match the one used internally.

Fix errors reported by flake8, black, and usort.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/24

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,13 +14,36 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v1
+      - uses: actions/setup-python@v2
 
-      - name: Install Dependencies
-        run: |
-          python -m pip install --upgrade pip
-          python -m pip install black==20.8b1
+      - name: Install black
+        run: pip install black==21.4b2
 
-      - name: Run Black
-        run: |
-          black --check --diff .
+      - name: Run black
+        run: black --check --diff .
+
+  usort:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+
+      - name: Install usort
+        run: pip install usort==0.6.4
+
+      - name: Run usort
+        run: usort diff . && usort check .
+
+  flake8:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+
+      - name: Install flake8
+        run: pip install flake8
+
+      - name: Run flake8
+        run: flake8
--- a/.flake8
+++ b/.flake8
@@ -22,4 +22,6 @@ exclude =
   _ig_fbcode_wheel/*,
   buck-out/*,
   third-party-buck/*,
-  third-party2/*
+  third-party2/*,
+  source/interprocedural_analyses/taint/test/integration/*,
+  source/command/test/integration/fake_repository/*
--- a/api/connection.py
+++ b/api/connection.py
@@ -10,7 +10,7 @@ import logging
 import subprocess
 from pathlib import Path
 from types import TracebackType
-from typing import Any, Dict, List, NamedTuple, Optional
+from typing import Any, List, NamedTuple, Optional
 
 from typing_extensions import TypedDict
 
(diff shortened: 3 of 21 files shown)


.github/workflows/lint.yml
```diff
@@ -14,13 +14,23 @@ jobs:
...
+        run: black --diff .
+
+  usort:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+
+      - name: Install black
```
0xedward, line 32:
```suggestion
      - name: Install usort
```

.github/workflows/lint.yml
```diff
@@ -14,13 +14,23 @@ jobs:
...
+        run: pip install usort==0.6.4
+
+      - name: Run black
```
0xedward, line 35:
```suggestion
      - name: Run usort
```

.github/workflows/lint.yml
```diff
@@ -14,13 +14,23 @@ jobs:
...
-          python -m pip install --upgrade pip
-          python -m pip install black==20.8b1
+      - name: Install black
+        run: pip install black==21.4b2
 
-      - name: Run Black
-        run: |
-          black --check --diff .
+      - name: Run black
+        run: black --diff .
```
0xedward, line 23:
```suggestion
        run: black --check --diff .
```

From my testing, it looks like running `black` with `--diff` will exit with 0 regardless if the file has been requires reformatting. Since we want Github Actions to [display a failure when changes are required](https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions), we still might need to include the [`--check`](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html?#exit-code) flag here

.github/workflows/lint.yml
```diff
@@ -14,13 +14,23 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+
+      - name: Install black
+        run: pip install usort==0.6.4
+
+      - name: Run black
+        run: usort diff .
```
0xedward, line 36:
```suggestion
        run: usort diff . && usort check .
```

Same reason (exit code) as adding `--check` flag to `black`

0xedward (changes requested):
Thanks for working on this :)

I added a few comments regarding exit codes. For example, the [most recent GitHub Actions run on this PR](https://github.com/facebook/pyre-check/pull/453/checks?check_run_id=3065776585) shows that `usort` would format `pyre_extensions/tests/simple_tests.py`, but the Github Actions job is displayed as successful

r0rshark (changes requested):
Yes internally we use flake8 a similar configuration to the internal one is defined in https://github.com/facebookresearch/FBPCS/blob/main/.flake8 . We can start with that and maybe exclude some folder like `./source/interprocedural_analyses/taint/test/integration/`. If there are still error we may create a pull request to fix those as well it may be possible that they have been suppressed internally but if it's not a huge effort we can address them :)

abishekvashok:
@0xedward thanks for the suggestions :) Fixed as per suggestions

abishekvashok:
@r0rshark okay, I included flake8 here.

There are issues that flake8 reports. I could have made another PR, but I feel it can be fixed here itself, so made those commits, 1 fixing flake8 issues, 1 fixing usort, and the other suppressing a black issue that conflicts with usort.

r0rshark:
@abishekvashok  Sorry I didn't notice we have already a custom .flake8 configuration... What happens if we keep it like it is currently and only add to the exclusion:
-   source/interprocedural_analyses/taint/test/integration/*,
-   source/command/test/integration/fake_repository/*
Do we still see many other issues raised by flake8?

abishekvashok:
@r0rshark 
> Do we still see many other issues raised by flake8?

Ya just a single issue about code logic that I don't want to touch because that might break something, maybe, but yeah let's see, the others I can solve.

abishekvashok:
@r0rshark sorry for the back and forth (commit log, and rewriting history), but it should be all set now! Thanks for the suggestions.

PR #454, opened 2021-07-15
Add DSL model generator for tornado RequestHandler

abishekvashok:
A way to define HTTPRequest handler with tornado is to subclass
tornad.web.RequestHandler and implement methods. Reviewed the API and
implements a DSL model generattor to automatically create UserControlled
models.

Fixes https://github.com/MLH-Fellowship/pyre-check/issues/29

### DIFF
--- a/stubs/third_party_taint/tornado_sources_sinks.pysa
+++ b/stubs/third_party_taint/tornado_sources_sinks.pysa
@@ -45,3 +45,22 @@ tornado.httputil.HTTPServerRequest.body: TaintSource[UserControlled, UserControl
 tornado.httputil.HTTPServerRequest.arguments: TaintSource[UserControlled, UserControlled_Payload]
 tornado.httputil.HTTPServerRequest.query_arguments: TaintSource[UserControlled, UserControlled_Payload]
 tornado.httputil.HTTPServerRequest.body_arguments: TaintSource[UserControlled, UserControlled_Payload]
+
+ModelQuery(
+  find = "methods",
+  where = [
+    parent.extends("tornado.web.RequestHandler", is_transitive=True),
+    AnyOf(
+      name.matches("\.get$"),
+      name.matches("\.post$"),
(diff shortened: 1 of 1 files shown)


abishekvashok:
Screenshot 2021-07-15 at 11 47 52 AM

stubs/third_party_taint/tornado_sources_sinks.pysa
```diff
@@ -45,3 +45,22 @@ tornado.httputil.HTTPServerRequest.body: TaintSource[UserControlled, UserControl
 tornado.httputil.HTTPServerRequest.arguments: TaintSource[UserControlled, UserControlled_Payload]
 tornado.httputil.HTTPServerRequest.query_arguments: TaintSource[UserControlled, UserControlled_Payload]
 tornado.httputil.HTTPServerRequest.body_arguments: TaintSource[UserControlled, UserControlled_Payload]
+
+ModelQuery(
+  find = "methods",
+  where = [
+    parent.extends("tornado.web.RequestHandler", is_transitive=True),
+    AnyOf(
+      name.matches("\.get+$"),
```
gbleaney, line 54: Why do we have the `+` here? That just makes the `t` in `get` optional:
image

We should probably remove it.

```suggestion
      name.matches("\.get$"),
```

gbleaney (changes requested):
Looks great, just one small question on the regex. Thanks for the PR, this adds some much needed coverage!

stubs/third_party_taint/tornado_sources_sinks.pysa
abishekvashok: Tbh, I am not so good in regex, I tried a generator just like the one in the pic 😅, I did try out few combinations but not ge 😅, it seems you are right! I will make the change!

abishekvashok:
@gbleaney the request change has been made! Thanks for the suggestion, really appreciate the work you put into the review :)

PR #455, opened 2021-07-17
Adds sinks for psycopg2

PulkitMishra:
Stubs for psycopg2 have recently been added to typesshed
https://github.com/python/typeshed/blob/a0f199727b68917a80939432e1e69315afc44608/stubs/psycopg2/psycopg2/_psycopg.pyi#L333

Since Postgres is a widely used database and psycopg its most popular adapter, this PR adds sinks for psycopg2 to pysa.


abishekvashok (approved):
Seems cool enough

onionymous:
Looks great! Would you mind adding a test plan showing the new models generated, or injecting an example issue and ensuring it can be caught with the new models? (e.g. using the deliberately vulnerable flask app).

PR #461, opened 2021-08-11
Updated tests scripts to run tests from one directory higher

m0mosenpai:
Related to discussion on this with @gbleaney. 

PR #433's use of relative imports causes `Import beyond top-level package` errors. Internally, relative imports are required to pass the tests whereas on GitHub, this causes issues. This PR fixes that issue by running tests from one directory higher, unblocking #433.

### DIFF
--- a/tools/pysa_integration_tests/utils.py
+++ b/tools/pysa_integration_tests/utils.py
@@ -155,7 +155,7 @@ def run_pysa_integration_test(
     if run_from_source:
         command = [
             "python",
-            "-m" "client.pyre",
+            "-m" "pyre-check.client.pyre",
             "--noninteractive",
             "analyze",
         ]
--- a/.github/workflows/pysa.yml
+++ b/.github/workflows/pysa.yml
@@ -40,7 +40,7 @@ jobs:
           unzip -qq ./stubs/typeshed/typeshed.zip -d ./typeshed/
           ./scripts/setup.sh --local --no-tests
           make -C source
-          echo "PYTHONPATH=$GITHUB_WORKSPACE:$PYTHONPATH" >> $GITHUB_ENV
+          echo "PYTHONPATH=$GITHUB_WORKSPACE/..:$PYTHONPATH" >> $GITHUB_ENV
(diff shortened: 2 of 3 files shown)


tools/pysa_integration_tests/utils.py
```diff
@@ -155,7 +155,7 @@ def run_pysa_integration_test(
     if run_from_source:
         command = [
             "python",
-            "-m" "client.pyre",
+            "-m" "pyre-check.client.pyre",
```
gbleaney, line 158: You also need to change the current working directory to make this work. See line 177

gbleaney (changes requested):

tools/pysa_integration_tests/utils.py
abishekvashok: Can I add in a suggestion here? Disclaimer- I don't know what the other PR is about. Sorry for coming in out of nowhere, but I know the tests are failing (pysa) because, we are adding the repository root to python path in the GitHub action:
https://github.com/facebook/pyre-check/blob/306428394eafd7cc90eab50c9d1bf8befdcb25a5/.github/workflows/pysa.yml#L43

So it checks client folder with pyre.py file in the root of the repository. This change might be un-needed. :)

gbleaney: @abishekvashok this is a great suggestion! If we update `PYTHONPATH` there to be one dir higher there, this will allow us to do the relative imports we need between `client` and `api`. We'll still need to keep the updated `-m` command, on this line, but we won't need to update the current working directory as I suggested

abishekvashok: Yes, I viewed the other PR - and now I understand this better :), updating the `PYTHONPATH` and `pythonLocation` in pysa.yml along with this diff should fix the GitHub action!

m0mosenpai: Ohh I see. Just to clarify, no other changes would be needed in this PR and the updation of `PYTHONPATH` will happen in the work @abishekvashok you guys are doing? Or shall I make those changes here?

gbleaney: @m0mosenpai you should update the PYTHONPATH in this PR

PR #464, opened 2021-08-16
Adds native sarif output support to Pyre

abishekvashok:
Adds native support to output pyre results in sarif format. Modifies
existing behaviour of outputing results to a json file and then using a
custom python script to generate sarif output to be within pyre itself,
so that the output can be pushed out in sarif format with --output=sarif
option.

This makes running Pyre in Github Action and other continous integration
tests or even localally to generate sarif results easier.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/50


abishekvashok:
See the sarif results produced when uploaded producing the analysis at: https://github.com/abishekvashok/pyre-check/pull/4

I don't know how you guys make the nightly releases, hence, I made two commits - one to add sarif output support, and the other to modify the GitHub actions. Since, the GitHub actions make use of pyre-check-nightly and building from source is very slow, we could merge the second commit when a nightly build with the changes in the first commit are landed.

If that is the case, let me know and I'll be happy to rebase :)

Testing screencast:

https://user-images.githubusercontent.com/8947010/129564580-00c825a6-803e-4830-bdb2-63bd1157ca5c.mov

Lint (flake8) is failing unrelated to this PR

client/error.py
```diff
@@ -240,6 +265,70 @@ def to_text(self) -> str:
...
+    return {
+        "version": "2.1.0",
+        "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.4.json",
+        "runs": [
+            {
+                "tool": {
+                    "driver": {
+                        "name": "Pyre",
+                        "informationUri": "https://www.pyre-check.org",
+                        "rules": [
```
0xedward, line 305: Not sure if this is currently possible, but it would be cool if we could list [specific codes for type errors](https://pyre-check.org/docs/errors/#error-codes)

https://github.com/facebook/pyre-check/blob/bfa69fa7198215e07fcdd20d6e81fc5a03dc0253/source/analysis/analysisError.ml#L422-L557

abishekvashok: @0xedward nice suggestion! Made the required changes. (Teaser: https://github.com/abishekvashok/pyre-check/security/code-scanning/3?query=ref%3Arefs%2Fpull%2F4%2Fmerge)

client/error.py
```diff
@@ -240,6 +265,76 @@ def to_text(self) -> str:
...
+            {
+                "id": "PYRE-ERROR-" + str(error.code)
+                if error.code is not None
+                else "PYRE-ERROR-MDL",
+                "name": name.title().replace(" ", ""),
+                "shortDescription": {"text": name},
+                "helpUri": "https://www.pyre-check.org",
+                "help": {"text": name},
+            }
+        )
```
grievejia, line 319: Looking at the context, I think it's more consistent to push these error-specific `rules` computation into individual error class themselves (just like how `results` were computed). E.g. either make `Error`, `LegacyError` and `ModelVerificationError` all have two methods `to_sarif_result` and `to_sarif_rule`, or have their `to_sarif` method return a dataclass that has both a result and a rule field.

grievejia (changes requested):
Style nit

client/error.py
abishekvashok: @grievejia thank you for your time and looking into this. You are right! Spot on - refactored as per suggestion by adding methods to all 3 error classes.

client/error.py
```diff
@@ -240,6 +277,71 @@ def to_text(self) -> str:
...
+    errors: Union[
+        Sequence[Error], Sequence[LegacyError], Sequence[ModelVerificationError]
+    ]
+) -> Dict[str, Any]:
+
+    results: List[Dict[str, Any]] = []
+    rules: List[Dict[str, Any]] = []
+    for error in errors:
+        results.append(error.to_sarif())
+        rules.append(error.get_sarif_rule())
```
grievejia, line 326: Cosmetic nit:
```
results = [ error.to_sarif() for error in errors ]
rules = [ error.get_sarif_rule() for error in errors ]
```

client/error.py
```diff
@@ -240,6 +277,71 @@ def to_text(self) -> str:
...
+    return {
+        "version": "2.1.0",
+        "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.4.json",
+        "runs": [
+            {
+                "tool": {
+                    "driver": {
+                        "name": "Pyre",
+                        "informationUri": "https://www.pyre-check.org",
+                        "rules": list({rule["id"]: rule for rule in rules}.values()),
```
grievejia, line 337: Is this trying to deduplicate rules by id? Would be nice to have a comment there.

grievejia (approved):
LGTM with minor nits

.github/workflows/pyre.yml
```diff
@@ -14,31 +14,20 @@ jobs:
...
-          python -m pip install --upgrade pip
-          python -m pip install -r requirements.txt
-          python -m pip install cython
-          python -m pip install flask flask_cors graphql-core typing_inspect
+          pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install cython flask flask_cors graphql-core typing_inspect
           VERSION=$(grep "version" .pyre_configuration | sed -n -e 's/.*\(0\.0\.[0-9]*\).*/\1/p')
-          python -m pip install pyre-check-nightly==$VERSION
+          pip install pyre-check-nightly==$VERSION
```
arthaud, line 25: Those formatting changes should probably have been in a different pull request. It doesn't seem related to your changes and just hinders code review.

arthaud (changes requested):
Could you rebase this on top of main? It looks like there might be merge conflicts.

client/error.py
abishekvashok: Yes! I will add a comment here.

.github/workflows/pyre.yml
abishekvashok: Thanks. I will keep this in mind :)

client/error.py
abishekvashok: Wouldn't that be `O(2n)` instead of `O(n)` as in the diff? :) I will make the change nonetheless

.github/workflows/pyre.yml
abishekvashok: @arthaud I added this as an additional commit for now, but if you want - I can move it to a new branch, wait the first commit to be merged, and then introduce the other two commits in separate PRs.

abishekvashok:
@arthaud @grievejia done :)

r0rshark:
@abishekvashok it looks like the Pyre action https://github.com/facebook/pyre-check/pull/464/checks?check_run_id=3379460464 is failing at the SARIF step because it tries to upload an empty file and then tries to parse it
```
With the provided path, there will be 1 file uploaded
Total size of all the files uploaded is 0 bytes
...
Error: Unexpected end of JSON input
  SyntaxError: Unexpected end of JSON input
```

abishekvashok:
@r0rshark thanks for taking a look ;)
yes it will fail here - because it uses nightly version and the sarif output isn't in the latest nightly version. It will pass when the latest nightly version gets released. However, I can move the actions change and sarif output into two PRS. Quote from me above

> I don't know how you guys make the nightly releases, hence, I made two commits - one to add sarif output support, and the other to modify the GitHub actions. Since, the GitHub actions make use of pyre-check-nightly and building from source is very slow, we could merge the second commit when a nightly build with the changes in the first commit are landed.

> If that is the case, let me know and I'll be happy to rebase and possibly move the second commit to another pull request.

client/error.py
```diff
@@ -240,6 +277,68 @@ def to_text(self) -> str:
...
+    errors: Union[
+        Sequence[Error], Sequence[LegacyError], Sequence[ModelVerificationError]
+    ]
+) -> Dict[str, Any]:
+
+    results: List[Dict[str, Any]] = [error.to_sarif() for error in errors]
+    rules: List[Dict[str, Any]] = [error.get_sarif_rule() for error in errors]
+    return {
+        "version": "2.1.0",
+        "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.4.json",
```
0xedward, line 326: I think this is link to the "official" schema, so let's use that instead

```suggestion
        "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
```

dkgi:
> yes it will fail here - because it uses nightly version and the sarif output isn't in the latest nightly version. It will pass when the latest nightly version gets released. However, I can move the actions change and sarif output into two PRS. Quote from me above

Sorry I missed this. Would strongly prefer if we could split this up so that we can do a nightly release (this gets pushed from FB infrastructure unfortunately) and then update the GitHub action in a second PR.

abishekvashok:
I will split and get back :)

PR #467, opened 2021-08-21
Adds model verification testing

PulkitMishra:
WIP - creates a model validation testing system for the Opensource models

### DIFF
--- a/tools/pysa_integration_tests/model_verify.py
+++ b/tools/pysa_integration_tests/model_verify.py
@@ -1,0 +1,30 @@
+import subprocess
+import venv
+import os
+import shutil
+import tempfile
+from pathlib import Path
+from ...api.connection import PyreConnection
+from ...api.query import get_invalid_taint_models
+
+cur_dir = Path(__file__).parent
+envpath = cur_dir / ".env"
+requirements_path = cur_dir / "verify_models" / "requirements.txt"
+activate = envpath / "bin" / "activate"
+command = f"pip install -r {requirements_path}"
+
+if os.path.exists(envpath):
+  shutil.rmtree(envpath)
+os.mkdir(envpath)
+venv.create(envpath, with_pip=True)
+
+
+subprocess.check_call(["bash", "-c", f"source {activate}; {command}"])
+
+pyre_connection = PyreConnection(cur_dir / "verify_models")
+
+temp_file = tempfile.TemporaryFile(prefix=".pyre_configuration", suffix=".local", dir=cur_dir)
+temp_file.write(b'{"source_directories": ["verify_models"]}')
+
+print(get_invalid_taint_models(pyre_connection))
+shutil.rmtree(envpath)
--- a/tools/pysa_integration_tests/verify_models/requirements.txt
+++ b/tools/pysa_integration_tests/verify_models/requirements.txt
@@ -1,0 +1,3 @@
+Flask
+lxml
+pyre-check
--- a/tools/pysa_integration_tests/verify_models/README.md
+++ b/tools/pysa_integration_tests/verify_models/README.md
@@ -1,0 +1,8 @@
+# Overview
+
+This folder helps model_verify.py in model validation.
+
+# Contents
+
+1. requirements.txt : File that contain the library dependencies needed for the models defined.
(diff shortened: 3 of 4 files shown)


tools/pysa_integration_tests/model_verify.py
```diff
@@ -0,0 +1,30 @@
...
+os.mkdir(envpath)
+venv.create(envpath, with_pip=True)
+
+
+subprocess.check_call(["bash", "-c", f"source {activate}; {command}"])
+
+pyre_connection = PyreConnection(cur_dir / "verify_models")
+
+temp_file = tempfile.TemporaryFile(prefix=".pyre_configuration", suffix=".local", dir=cur_dir)
+temp_file.write(b'{"source_directories": ["verify_models"]}')
```
abishekvashok, line 27: Put in as well taint models path pointing to the repositories stubs? This helps us make the action verify newer models as well.

tools/pysa_integration_tests/verify_models/requirements.txt
```diff
@@ -0,0 +1,3 @@
+Flask
```
abishekvashok, line 1: I am sorry to say it, but we may want to add all the python modules or libraries for which we have stubs here.

tools/pysa_integration_tests/model_verify.py
```diff
@@ -0,0 +1,30 @@
...
+
+print(get_invalid_taint_models(pyre_connection))
```
abishekvashok, line 29: What output do you get? Can we have a GitHub action that runs this script?

tools/pysa_integration_tests/model_verify.py
```diff
@@ -0,0 +1,30 @@
+import subprocess
```
abishekvashok, line 1: Fb copyright notice to all files please :)

abishekvashok (changes requested):

tools/pysa_integration_tests/model_verify.py
PulkitMishra: yes having a github action is the next step @abishekvashok but first need to find a way to make pyre_configuration.local play well with pyre_configuration at the root folder

abishekvashok: If doesn't play well, feel free to delete the .pyre_configuration file at the root of the dir and create a file at the your folder. For an action it should get you by :)

PulkitMishra: yeah that's possible and is my last resort. Some clever use of tempfile seems cleaner though

r0rshark: @PulkitMishra  To avoid the issue with the .pyre_configuration located in the root folder the trick is to create a temporary directory in the global temporary folder (with something like `tempfile.gettempdir()` ) and create a folder there in which we can store the .pyre_configuration file. We can then pass to the PyreConnection the path of this folder so it will not conflict with the global .pyre_configuration. 
We can create a function which does this in https://github.com/facebook/pyre-check/blob/main/api/connection.py and invoke it passing `tempfile.gettempdir()` as root directory and the path to the virtualenvironment we created as a source dir. Example function:
```
def create_temporary_pyre_connection(
    root_directory: pathlib.Path
    targets: Optional[List[str]] = None,
    source_directories: Optional[List[str]] = None,
) -> PyreConnection:
  # We can compute the sha1 as folder name to avoid creating a new directory every time and use the same if the source 
  directories and targets do not change
  root_directory.mkdir(exist_ok=True)
  pyre_configuration_directory = (
          root_directory
          / hashlib.sha1(
              ",".join((targets or []) + (source_directories or [])).encode()
          ).hexdigest()
      ) 
    pyre_configuration_directory.mkdir(exist_ok=True)
    with open(
        str(pyre_configuration_directory / ".pyre_configuration.local"), "w"
    ) as configuration:
        configuration_contents: Dict[str, Any] = {}
        if targets:
            configuration_contents["targets"] = targets
        if source_directories:
            configuration_contents["source_directories"] = source_directories
        json.dump(configuration_contents, configuration, indent=2)
    return PyreConnection(pyre_configuration_directory)
```

PR #508, opened 2021-10-13
Adds more details to ModelVerification error

abishekvashok:
Adds more details to model verification error by adding additional
pieces of code that wrap over the errors passed to the python from the
Ocaml code and see whether the module in the error is installed or not
using pkg_resources from setuptools.

This implementation is kinda hacky and is dependentant on the error
message (description) format of the ModelVerification errors. A suitable
way to tackle this would be to pass the module name as well from Ocaml.

Before the changes:
Before

After the changes:
After

Test plan:
 - Create a new virtual env
 - Build pyre-check from source using instructions in https://pyre-check.org/docs/installation/#building-from-source
 - Run pyre analyze (make sure pyre is the locally built module not the one installed from pip)
 - See the additional information in model verification errors.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/81

### DIFF
--- a/client/commands/v2/validate_models.py
+++ b/client/commands/v2/validate_models.py
@@ -9,6 +9,8 @@ import os
 from pathlib import Path
 from typing import List, Dict
 
+import pkg_resources
+
 from ... import commands, configuration as configuration_module, error as error_module
 from . import query, server_connection, remote_logging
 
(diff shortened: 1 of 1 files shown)


abishekvashok:
Closing this for a better Ocaml version!

PR #522, opened 2021-11-03
Adds AllOf constraint to Pysa DSL

abishekvashok:
Adds support for a new constraint - `AllOf` to Pysa DSL, to be used in
the where clause. This query would find methods which obey all of the
conditions specified inside it and then model only such methods.

Screenshot:
Screenshot 2021-11-03 at 8 23 44 PM

Test Plan
- create a new model query with the new `AllOf` constraint.
- run `pyre analyse` with `dump-model-results-to` after compiling the source (from this branch)
- See the constraint in action. Maybe, play with it a bit...

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/86


source/interprocedural_analyses/taint/test/integration/model_query.py.pysa
```diff
@@ -124,6 +124,18 @@ ModelQuery(
   ]
 )
 
+ModelQuery(
```
onionymous, line 127: Hmm, you'll need to run the integration tests to get the generated changes after updating the integration test here. Your changes to model_query.py should be included, in addition to the generated changes to model_query.py.models, model_query.py.cg, model_query.py.overrides, in this PR

onionymous (changes requested):
Thanks for doing this! :)

source/interprocedural_analyses/taint/test/integration/model_query.py.pysa
abishekvashok: Hey @onionymous 
> I haven't run the integration tests yet, because I don't know how to run them (Running python3 ./scripts/run_integration_tests.sh ./ for instance says that ~/pyre-check/.DS_Store is not directory!

abishekvashok:
@onionymous  done :)

source/interprocedural_analyses/taint/test/integration/model_query.py
```diff
@@ -236,3 +236,13 @@ def method1():
...
+
+
+class ClassTest3_Alarm1:
+    def method1():
+        return 0
+
+    def method2():
+        return 0
+
+
```
onionymous, line 248: lint: extra blank line here

onionymous (changes requested):

abishekvashok:
Wow that was fast!

PR #526, opened 2021-11-04
Modify existing pyre-check action to use newly added native sarif support

abishekvashok:
Modifies existing pyre-check github action to use the newly added sarif
output support. Also remove redundant scripts.

Aftermath of a new pyre-check version being released on pypi with sarif support!

Test plan:
See the GitHub action!


.github/workflows/pyre.yml
```diff
@@ -14,31 +14,20 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v1
+      - uses: actions/setup-python@v2
 
-      - name: Install Dependencies
+      - name: Install Dependencies and pyre-check
```
dkgi, line 19: Why the change here? I think of this as "dependencies for the action" and Pyre is just a normal dependency.

abishekvashok: I wanted to give it some special status but yeah it's fine either way :)

abishekvashok:
@dkgi thanks for the review, removed the unnecessary change :)

PR #529, opened 2021-11-05
Adds key validation to taint.config files

abishekvashok:
Adds validation of required and optional keys of taint.config files to
taintConfiguration parser in the server.

Previosuly, we didn't do such validation and ignored invalid keys in the
file and errored out with "not so pretty" error message when a required
key was absent.

Now we error out with a pretty error message and show a warning about
invalid keys while continuing through the parsing process.

Screenshot:
Screenshot 2021-11-05 at 11 30 47 PM

Test Plan
- Compile the server from source
- add an (or some) invalid key(s) to your taint.config file
- run pysa with PYRE_BINARY pointing to the executable build from source
- see the invalid key message
- remove a (or some) required key(s) from your taint.config file
- run pysa with PYRE_BINARY pointing to the executable build from source
- see the pretty new error.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/87

### DIFF
--- a/source/interprocedural_analyses/taint/taintConfiguration.ml
+++ b/source/interprocedural_analyses/taint/taintConfiguration.ml
@@ -223,14 +223,35 @@ let parse source_jsons =
     | `String "parametric" -> AnnotationParser.Parametric
     | unexpected -> failwith (Format.sprintf "Unexpected kind %s" (Json.Util.to_string unexpected))
   in
+  let check_keys ~required_keys ~valid_keys ~current_keys ~section =
+    let valid_keys_hash_set = String.Hash_set.of_list valid_keys in
+    let current_keys_hash_set = String.Hash_set.of_list current_keys in
+    let check_required_key_present key =
+      if not (Hash_set.mem current_keys_hash_set key) then
+        failwith (Format.sprintf "Required key `%s` is not found in section `%s`" key section)
+    in
+    let check_key_is_valid key =
+      if not (Hash_set.mem valid_keys_hash_set key) then (
+        Log.error "Unknown key `%s` encountered in section `%s`" key section)
+    in
+    List.iter current_keys ~f:check_key_is_valid;
+    List.iter required_keys ~f:check_required_key_present
+  in
   let parse_lineage_analysis json = json_bool_member "lineage_analysis" json ~default:false in
   let parse_string_list json = Json.Util.to_list json |> List.map ~f:Json.Util.to_string in
-  let parse_source_or_sink json =
+  let parse_source_or_sink section json =
+    check_keys
+      ~required_keys:["name"]
+      ~current_keys:(Json.Util.keys json)
+      ~valid_keys:["name"; "comment"]
+      ~section;
     let name = Json.Util.member "name" json |> Json.Util.to_string in
     { AnnotationParser.name; kind = kind json }
   in
-  let parse_sources json = array_member "sources" json |> List.map ~f:parse_source_or_sink in
-  let parse_sinks json = array_member "sinks" json |> List.map ~f:parse_source_or_sink in
+  let parse_sources json =
+    array_member "sources" json |> List.map ~f:(parse_source_or_sink "sources")
+  in
+  let parse_sinks json = array_member "sinks" json |> List.map ~f:(parse_source_or_sink "sinks") in
   let parse_features json =
     let parse_feature json = Json.Util.member "name" json |> Json.Util.to_string in
     array_member "features" json |> List.map ~f:parse_feature
@@ -243,6 +264,12 @@ let parse source_jsons =
   in
   let parse_rules ~allowed_sources ~allowed_sinks json =
     let parse_rule json =
+      let required_keys = ["name"; "code"; "sources"; "sinks"; "message_format"] in
+      check_keys
+        ~required_keys
+        ~valid_keys:required_keys
+        ~current_keys:(Json.Util.keys json)
+        ~section:"rules";
       let sources =
         Json.Util.member "sources" json
         |> parse_string_list
@@ -336,14 +363,13 @@ let parse source_jsons =
               code;
               message_format;
             }
-            ::
-            {
-              Rule.sources = second_sources;
-              sinks = List.map second_sinks ~f:(fun sink -> Sinks.TriggeredPartialSink sink);
-              name;
-              code;
-              message_format;
-            }
+            :: {
+                 Rule.sources = second_sources;
+                 sinks = List.map second_sinks ~f:(fun sink -> Sinks.TriggeredPartialSink sink);
+                 name;
+                 code;
+                 message_format;
+               }
             :: rules,
             PartialSinkConverter.add
               partial_sink_converter
@@ -363,6 +389,11 @@ let parse source_jsons =
     match member "implicit_sinks" json with
     | `Null -> empty_implicit_sinks
     | implicit_sinks ->
+        check_keys
+          ~required_keys:[]
+          ~valid_keys:["conditional_test"; "literal_strings"]
+          ~current_keys:(Json.Util.keys implicit_sinks)
+          ~section:"implicit_sinks";
         let conditional_test =
           match member "conditional_test" implicit_sinks with
           | `Null -> []
@@ -396,6 +427,11 @@ let parse source_jsons =
     match member "implicit_sources" json with
     | `Null -> { literal_strings = [] }
     | implicit_sources ->
+        check_keys
+          ~required_keys:[]
+          ~valid_keys:["conditional_test"; "literal_strings"]
+          ~current_keys:(Json.Util.keys implicit_sources)
+          ~section:"implicit_sources";
         let literal_strings =
           array_member "literal_strings" implicit_sources
           |> List.map ~f:(fun json ->
@@ -412,7 +448,6 @@ let parse source_jsons =
         in
         { literal_strings }
   in
-
   let sources = List.concat_map source_jsons ~f:parse_sources in
   let sinks = List.concat_map source_jsons ~f:parse_sinks in
   let features = List.concat_map source_jsons ~f:parse_features in


abishekvashok:
PS: The error message is appearing twice even though I am printing it only once. This must be in some way related to the fact that most error messages by pyre (like a package defined in search path is not present in the environment get's printed twice?)

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -223,9 +223,35 @@ let parse source_jsons =
...
   in
+  let unknown_keys = String.Hash_set.create () in
+  let check_keys ~required_keys ~valid_keys ~current_keys ~section =
+    let valid_keys_hash_set = String.Hash_set.of_list valid_keys in
+    let current_keys_hash_set = String.Hash_set.of_list current_keys in
+    let check_required_key_present key =
+      Hash_set.mem current_keys_hash_set key
+      |> function
+      | false ->
+          failwith (Format.sprintf "Required key %s is not found in section `%s`" key section)
```
arthaud, line 234: We should probably return a proper error instead of throwing an exception. That could be a follow-up.

arthaud, line 234: It's more natural as a if condition I believe
```suggestion
      if not (Hash_set.mem current_keys_hash_set key) then
          failwith (Format.sprintf "Required key %s is not found in section `%s`" key section)
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -223,9 +223,35 @@ let parse source_jsons =
...
+      | true -> ()
+    in
+    let check_key_is_valid key =
+      Hash_set.mem valid_keys_hash_set key
+      |> function
+      | false ->
+          Hash_set.add unknown_keys (Format.sprintf "Unknown key `%s` in section `%s`" key section)
```
arthaud, line 241: Could we just log the error here? I'm not a fan of this `unknown_keys` variable. It's easy to forget.

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -336,14 +368,13 @@ let parse source_jsons =
...
-              code;
-              message_format;
-            }
+            :: {
+                 Rule.sources = second_sources;
+                 sinks = List.map second_sinks ~f:(fun sink -> Sinks.TriggeredPartialSink sink);
+                 name;
+                 code;
+                 message_format;
+               }
```
arthaud, line 377: Hmm our auto-formatter disagrees with this. Are we using the same one?

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -396,6 +432,11 @@ let parse source_jsons =
     match member "implicit_sources" json with
     | `Null -> { literal_strings = [] }
     | implicit_sources ->
+        check_keys
+          ~required_keys:[]
+          ~valid_keys:["conditional_test"; "literal_strings"]
+          ~current_keys:(Json.Util.keys (Json.Util.member "implicit_sinks" json))
```
arthaud, line 438:
```suggestion
          ~current_keys:(Json.Util.keys implicit_sources)
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -363,6 +394,11 @@ let parse source_jsons =
     match member "implicit_sinks" json with
     | `Null -> empty_implicit_sinks
     | implicit_sinks ->
+        check_keys
+          ~required_keys:[]
+          ~valid_keys:["conditional_test"; "literal_strings"]
+          ~current_keys:(Json.Util.keys (Json.Util.member "implicit_sinks" json))
```
arthaud, line 400:
```suggestion
          ~current_keys:(Json.Util.keys implicit_sinks)
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -412,7 +453,15 @@ let parse source_jsons =
         in
         { literal_strings }
   in
-
+  let check_json json =
```
arthaud, line 456: nitpick: naming
```suggestion
  let check_json_keys json =
```

arthaud (changes requested):

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -466,6 +515,9 @@ let parse source_jsons =
   let lineage_analysis = List.exists ~f:parse_lineage_analysis source_jsons in
   let rules = List.rev_append rules generated_combined_rules in
   let matching_sources, matching_sinks = matching_kinds_from_rules rules in
+  if not (Hash_set.is_empty unknown_keys) then (
+    Log.error "The following unknown keys were found in taint.config file(s): ";
+    Hash_set.iter unknown_keys ~f:(Log.error "%s"));
```
onionymous, line 520:
```suggestion
    Log.error "The following unknown keys were found in the taint.config file(s) and will be ignored: %s" 
    (Hash_set.to_list unknown_keys |> String.concat ~sep:", ")
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -412,7 +453,15 @@ let parse source_jsons =
         in
         { literal_strings }
   in
-
+  let check_json json =
+    check_keys
+      ~required_keys:["sources"; "sinks"; "features"; "rules"]
```
onionymous, line 458: I don't think the root taint.config has any required keys, actually. Features definitely aren't required, and we have some hardcoded sources/sinks/rules for testing so this shouldn't be necessary either.

onionymous (changes requested):
Thanks for doing this! There seem to be quite a lot of test failures on this. Please run `make test` in the `source/` folder and resolve the failures. A lot of them are to do with `features` key being required, just at a glance, and some other cases of error messages being a bit different from what is expected.

source/interprocedural_analyses/taint/taintConfiguration.ml
onionymous: `Json.Util.member` will throw an exception on type error (https://github.com/ocaml-community/yojson/blob/master/lib/util.mli#L61) so this does seem like the better way to go.

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Yes!

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: is it okay if I make this an asset bool like we do for model_verification errors?🤔

abishekvashok:
@arthaud @onionymous thanks for the comments. Modified as per suggestions :)

abishekvashok:
Tests in the GitHub Actions were failing before the changes, so is lint. Verified the server tests with `make test`.

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -223,14 +223,37 @@ let parse source_jsons =
     | `String "parametric" -> AnnotationParser.Parametric
     | unexpected -> failwith (Format.sprintf "Unexpected kind %s" (Json.Util.to_string unexpected))
   in
+  let unknown_keys_count = ref 0 in
+  let check_keys ~required_keys ~valid_keys ~current_keys ~section =
+    let valid_keys_hash_set = String.Hash_set.of_list valid_keys in
+    let current_keys_hash_set = String.Hash_set.of_list current_keys in
+    let check_required_key_present key =
+      if not (Hash_set.mem current_keys_hash_set key) then
+        failwith (Format.sprintf "Required key %s is not found in section `%s`" key section)
```
arthaud, line 232: nitpick: add backticks
```suggestion
        failwith (Format.sprintf "Required key `%s` is not found in section `%s`" key section)
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -466,6 +503,10 @@ let parse source_jsons =
   let lineage_analysis = List.exists ~f:parse_lineage_analysis source_jsons in
   let rules = List.rev_append rules generated_combined_rules in
   let matching_sources, matching_sinks = matching_kinds_from_rules rules in
+  if !unknown_keys_count != 0 then
+    Log.error
+      "%d unknown keys were found in the taint.config file(s) and will be ignored."
+      !unknown_keys_count;
```
arthaud, line 509: What is the objective with this additional message? We already say `Unknown key X encountered in section Y`. Could we just say that it will be ignored there? I would try to avoid mutable variables if possible.

arthaud (changes requested):

source/interprocedural_analyses/taint/taintConfiguration.ml
arthaud: Could we add this back but without the required keys? It's still useful to know if we have unknown keys at the root.

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: >  I would try to avoid mutable variables if possible.
Sure let's remove it :)

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Is backticks for values from users?

arthaud: I guess you have a point. I'm fine either way.

abishekvashok:
I can and will squash this :)

abishekvashok:
Squashed :)

PR #532, opened 2021-11-10
Migrate pysa playground backend to websockets

abishekvashok:
Migrate pysa playground flask endpoints to websocket connections instead
of http connection. This enables us to provide the user with live
information about a particular run.

Test Plan
- stack #533 on top of this PR
- run the backend: `cd tools/sandbox && python3 application.py`
- run the frontend: `cd documentation/website && yarn start`
- goto the playground: http://locahost:3000/pysa_playground
- play around :)

Screencast:

https://user-images.githubusercontent.com/8947010/141153869-f476ad0c-285d-4993-a62a-d12cdd2601d6.mov

### DIFF
--- a/tools/sandbox/application.py
+++ b/tools/sandbox/application.py
@@ -17,6 +17,7 @@ from typing import IO, List
 
 from flask import Flask, request, jsonify
 from flask_cors import CORS
+from flask_socketio import SocketIO, emit
 
 logging.basicConfig(
     format="%(asctime)s %(levelname)s %(message)s",
@@ -111,7 +112,9 @@ class Pyre:
 
 
 class Pysa:
-    def __init__(self, model: str = "", use_builtin_pysa_models: bool = False) -> None:
+    def __init__(
+        self, input: str, model: str = "", use_builtin_pysa_models: bool = False
+    ) -> None:
         self._directory: Path = Path(tempfile.mkdtemp())
         self._stubs: Path = Path(tempfile.mkdtemp())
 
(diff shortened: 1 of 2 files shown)


abishekvashok:
**Pre-submission checklist**
    - [x] `pre-commit run`

tools/sandbox/application.py
```diff
@@ -188,39 +208,36 @@ def check() -> str:
...
-        or request.json.get("use_builtin_pysa_models")
-    )
-    model = (
-        request.args.get("model")
-        or request.form.get("model")
-        or request.json.get("model")
-    )
+@socketio.on("analyze", namespace="/analyze")
+def analyze(json) -> None:
+    print(json.get("input", None))
```
dkgi, line 213: Presumably we don't want to be printing to stdout here? Is this leftover from debugging?

dkgi (changes requested):
For my own education: what's the tl;dr on why websockets over http requests?

tools/sandbox/application.py
abishekvashok: Yeah, I do a bit of hard testing and this is the remainders from such a debugging process. So sorry.

abishekvashok:
> what's the tl;dr on why websockets over http requests?

@dkgi The pysa sandbox takes a bit of time to run Pysa. So there's a long delay of clicking the `Run Pysa` button and getting all the results from the backend. With web sockets, we can get results incrementally from the backend server. 

Rest is tl;dr

Typical http connections are non-persistent. Even though HTTP does support persistent connections, the client has to make a request to get a response from the server as you know. With the web sockets we can actually mimic as if we are running the Pysa natively. There has been instances however where I figured we might throttle the connection by sending in large numb er of responses to the client in a short span of time, (eg: model verification errors), so in these scenarios we club the lines together and send a single big response (since the time difference of the next line would almost be identical) which doesn't throttle.

Also, this is not going to make the pysa sandbox idle. I am guessing that we should do more work here to ensure proper sessions so that we won't let a user spawn multiple threads etc. but that seems to be something we can do in another PR after this one.

PR #544, opened 2021-11-24
Make malformed configuration exceptions typed

abishekvashok:
A malformed configuration exception occurs when either the taint.config
file is not found or it is not a valid json file. During the validation
of taint.config file with the json specs, we raised an exception if it
was illegal.

Now added code to pretty print the exception via Log.error and specify
the line number and column number along with the file name.

Since YoJSON doesn't expose or store line numbers, this is implemented
by creating a custom Bi_output buffer and specifying it to YoJSON so
that YoJSON uses it for parsing and the last line of the contents would
be the line in which the error occured and the position and length of
such the last line would give us the line number and column number of
the error.

Test Plan
- Compile the source from scratch
- Malform a `taint.config` file by deleting a `,` or in some other way
- Run `pyre analyze` (with `$PYRE_BINARY` pointing to the newly built binary)
- See the new error instead of an exception
- Restore `taint.config` to initial version
- Run `pyre analyze` again to make sure nothing is broken with this change :)

Before this change:
```
ƛ Pyre encountered an internal exception: ("Taint__TaintConfiguration.MalformedConfiguration(\"/Users/abishekvashok/projects/stubs/stubs/taint/core_privacy_security/taint.config\", \"Line 7, bytes 4-37:\\nExpected ',' or '}' but found '{\\n      \\\"name\\\": \\\"Demo\\\",\\n      \\\"co'\")")
```

After this change:
Screenshot 2021-11-24 at 4 50 48 PM

Fixes part of: https://github.com/MLH-Fellowship/pyre-check/issues/82

### DIFF
--- a/source/interprocedural_analyses/taint/taintConfiguration.ml
+++ b/source/interprocedural_analyses/taint/taintConfiguration.ml
@@ -721,21 +721,35 @@ let create
   =
   let file_paths = Path.get_matching_files_recursively ~suffix:".config" ~paths:taint_model_paths in
   let parse_configuration config_file =
+    try
     if not (Path.file_exists config_file) then
       raise
         (MalformedConfiguration { path = Path.absolute config_file; parse_error = "File not found" })
     else
-      try
+      let config_string =
         config_file
         |> File.create
         |> File.content
-        |> Option.value ~default:""
-        |> Json.from_string
+        |> Option.value ~default:"" in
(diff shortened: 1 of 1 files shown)


abishekvashok:
Just heard that `Bi_outbuf` would be invalidated in the next release of YoJSON :/

I am thinking of closing this

abishekvashok:
@0xedward closing due to above comment :)

PR #548, opened 2021-11-30
Filter results from backend for pysa playground

arpancodes:
**Pre-submission checklist**
    - [x] `pre-commit run`

## Summary

This PR solves the issue of sending all the information to the pysa playground client as result. Instead, now the backend only sends the line with either `INFO` or `ERROR` in it.

### DIFF
--- a/tools/sandbox/application.py
+++ b/tools/sandbox/application.py
@@ -152,18 +152,9 @@ class Pysa:
             text=True,
         ) as process:
             model_verification_errors = []
-            model_verification_errors_found = False
             for line in iter(process.stderr.readline, b""):
                 line = line.rstrip()
-                if line == "":
-                    break
-                elif "INFO" in line or "ERROR" in line:
-                    emit("pysa_results_channel", {"type": "output", "line": line})
-                elif "ERROR" in line and "is not part of the environment" in line:
-                    model_verification_errors.append(line)
-                    model_verification_errors_found = True
-                    continue
-                elif model_verification_errors_found:
+                if len(model_verification_errors) > 0:
                     # Emit all model verification lines together to prevent
                     # network overhead.
                     emit(
@@ -174,6 +165,12 @@ class Pysa:
                         },
                     )
                     LOG.debug("\n".join(model_verification_errors))
+                elif line == "":
+                    break
+                elif "ERROR" in line and "is not part of the environment" in line:
+                    model_verification_errors.append(line)
+                elif "INFO" in line or "ERROR" in line:
+                    emit("pysa_results_channel", {"type": "output", "line": line})
                 LOG.debug(line)
 
             return_code = process.wait()
--- a/tools/sandbox/application.py
+++ b/tools/sandbox/application.py
@@ -159,6 +159,9 @@ class Pysa:
                     break
                 elif "DEBUG" in line:
                     continue
+                elif "INFO" in line or "ERROR" in line:
+                    emit("pysa_results_channel", {"type": "output", "line": line})
+                    continue
                 elif "ERROR" in line and "is not part of the environment" in line:
                     model_verification_errors.append(line)
                     model_verification_errors_found = True
@@ -175,8 +178,6 @@ class Pysa:
                     )
                     LOG.debug("\n".join(model_verification_errors))
                 LOG.debug(line)
-                if "INFO" in line or "ERROR" in line:
-                    emit("pysa_results_channel", {"type": "output", "line": line})
 
             return_code = process.wait()
             if return_code != 0:
--- a/tools/sandbox/application.py
+++ b/tools/sandbox/application.py
@@ -157,11 +157,8 @@ class Pysa:
                 line = line.rstrip()
                 if line == "":
                     break
-                elif "DEBUG" in line:
-                    continue
                 elif "INFO" in line or "ERROR" in line:
                     emit("pysa_results_channel", {"type": "output", "line": line})
-                    continue
                 elif "ERROR" in line and "is not part of the environment" in line:
                     model_verification_errors.append(line)
                     model_verification_errors_found = True
--- a/tools/sandbox/application.py
+++ b/tools/sandbox/application.py
@@ -159,17 +159,20 @@ class Pysa:
                 elif "ERROR" in line and "is not part of the environment" in line:
                     model_verification_errors.append(line)
                 elif "INFO" in line or "ERROR" in line:
-                    if any(model_verification_errors):
+                    if model_verification_errors:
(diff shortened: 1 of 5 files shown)


tools/sandbox/application.py
```diff
@@ -175,7 +175,8 @@ def analyze(self) -> None:
                     )
                     LOG.debug("\n".join(model_verification_errors))
                 LOG.debug(line)
-                emit("pysa_results_channel", {"type": "output", "line": line})
+                if "INFO" in line or "ERROR" in line:
```
abishekvashok, line 178: Hey a simple suggestion would be to merge this with the previous ` if` statements. That would give us the opportunity to make the code a bit tidier. This approach works though - just a suggestion :)

arpancodes: made the changes 😄

abishekvashok (commented):
thanks for working on this, but wouldn't that continue prevent us from logging the line?
Also won't this change make the  preceding `if` clause, a line before unnecessary?

abishekvashok (commented):
LGTM

0xedward (changes requested):
Thanks for working on this @arpancodes and thanks for reviewing it @abishekvashok! 

This looks good to me, but I just noticed that we are using `model_verification_errors_found` to check if we should enter the following `elif`:
https://github.com/facebook/pyre-check/blob/72b268949606cd9a989ec22e3350cf32cba4afaa/tools/sandbox/application.py#L166-L174

Would we be able to remove the boolean entirely and check if `model_verification_errors` is empty instead? If so, since it's a small change maybe we could include it in this PR

tools/sandbox/application.py
```diff
@@ -157,8 +157,8 @@ def analyze(self) -> None:
                 line = line.rstrip()
                 if line == "":
                     break
-                elif "DEBUG" in line:
-                    continue
+                elif "INFO" in line or "ERROR" in line:
+                    emit("pysa_results_channel", {"type": "output", "line": line})
                 elif "ERROR" in line and "is not part of the environment" in line:
```
gbleaney, line 162: The body of this `elif` is now unreachable. The above `elif` will be entered if `line` contains `ERROR`, so we'll never make it into this one that is also checking for `line` to contain `ERROR`.

gbleaney (changes requested):

tools/sandbox/application.py
```diff
@@ -175,7 +175,6 @@ def analyze(self) -> None:
                     )
                     LOG.debug("\n".join(model_verification_errors))
                 LOG.debug(line)
```
arthaud, line 177: Pre-existing: This is never reached if we go through line 165 `continue`. I would put this at the beginning of the if/else chain, to prevent this type of confusion.
I would also remove that `continue`.

arthaud (changes requested):

tools/sandbox/application.py
arpancodes: My bad, reordered now.

tools/sandbox/application.py
arpancodes: Reordered and removed `continue` 😄

abishekvashok (commented):
Sorry for missing on the previous review. But I think now we would emit the model verification errors for every line after we find one.

abishekvashok:
I think something like this would work:
```python3
        model_verification_errors = []
            for line in iter(process.stderr.readline, b""):
                line = line.rstrip()

                if line == "":
                    break
                elif "ERROR" in line and "is not part of the environment" in line:
                    model_verification_errors.append(line)
                elif "INFO" in line or "ERROR" in line:
                    if any(model_verification_errors):
                        # Emit all model verification lines together to prevent
                        # network overhead.
                        emit(
                            "pysa_results_channel",
                        {
                            "type": "output",
                            "line": "\n".join(model_verification_errors),
                        },
                        )
                         
                    
                        LOG.debug("\n".join(model_verification_errors))
                        model_verification_errors = []
                    emit("pysa_results_channel", {"type": "output", "line": line})
                LOG.debug(line)
```

arpancodes:
Hey, @abishekvashok @0xedward since we are trying to emit all the model verification errors at once to avoid network overhead... should we add this one out of the loop? so it doesn't emit on every line but at once?

abishekvashok:
> since we are trying to emit all the model verification errors at once to avoid network overhead... should we add this one out of the loop? so it doesn't emit on every line but at once?

It doesn't emit now (before this diff). We can still make it not emit in a loop. This is due to the fact that as in the code snippet above, else if matches only once. 

If we add it outside the loop, I think we can only send after all other messages, or we would miss other messages from the server! If there is another way to push the message in the same order,  that'll be awesome but if not can you please try something along the lines I commented above?

tools/sandbox/application.py
```diff
@@ -152,30 +152,27 @@ def analyze(self) -> None:
...
-                    emit(
-                        "pysa_results_channel",
-                        {
-                            "type": "output",
-                            "line": "\n".join(model_verification_errors),
-                        },
-                    )
-                    LOG.debug("\n".join(model_verification_errors))
+                elif "INFO" in line or "ERROR" in line:
+                    if any(model_verification_errors):
```
0xedward, line 162: Nit: Is there a reason why we are using `any` instead of `len(model_verification_errors)` here?

```suggestion
                    if len(model_verification_errors) > 0:
```

tools/sandbox/application.py
```diff
@@ -152,30 +152,27 @@ def analyze(self) -> None:
...
+                        # Emit all model verification lines together to prevent
+                        # network overhead.
+                        emit(
+                            "pysa_results_channel",
+                            {
+                                "type": "output",
+                                "line": "\n".join(model_verification_errors),
+                            },
+                        )
+                        LOG.debug("\n".join(model_verification_errors))
```
0xedward, line 172: Should we create a variable to store the result of `"\n".join(model_verification_errors)` so we don't have to join twice?

```suggestion
                        model_verification_error_output = "\n".join(model_verification_errors)
                        emit(
                            "pysa_results_channel",
                            {
                                "type": "output",
                                "line": model_verification_error_output,
                            },
                        )
                        LOG.debug(model_verification_error_output)
```

0xedward (changes requested):

tools/sandbox/application.py
arpancodes: Hey yes! So @abishekvashok sent the snippet and it seemed to be working so it slipped by me... I'll remove it...

tools/sandbox/application.py
arpancodes: Correct, I'll refactor it.

PR #549, opened 2021-12-01
Adds line number support to taint.config parser

abishekvashok:
Adds line number support to taint.config parser to report line numbers
along with the validation errors and hence make the errors typed.
YoJSON, the currently used JSON persing module doesn't support line
number parsing, hence resorted to using another low level module -
JSONM.

However, JSONM lacks the high-level features YoJSON provides and is
pretty bare-bone. It would be a mundane task to rewrite code to include
only JSONM and ditch YoJSON. Hence, using YoJSON along with JSONM to
derive line numbers. Instead of storing the line numbers along during
parsing, we deligate the validation step to JSONM.

The validation step required a pass through the entire taint.config
files already, so migrating it to JSONM and providing only line numbers
for some errors won't be a big performance overhead.

Further, this approach enables us to use the 31% speed advantage YoJSON
provides over JSONM (line number parsing takes time) and by storing line
numbers only during errors, we can make efficient use of memory by
storing only the things we want!

A downside of this approach is that we would indeed be using two JSON
parsers!


abishekvashok:
Facing these weird compile time errors:
```
dune build @install -j auto --profile dev
File "interprocedural_analyses/taint/taintConfiguration.ml", line 560, characters 10-12:
560 |           in
                ^^
Error: Syntax error: ')' expected
File "interprocedural_analyses/taint/taintConfiguration.ml", line 554, characters 33-34:
554 |         | location_list::rest -> (
                                       ^
  This '(' might be unmatched
make: *** [dev] Error 1 
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -522,14 +523,46 @@ let parse source_jsons =
...
+          | [((starting_row, starting_column), (ending_row, ending_column))]
+          | ((starting_row, starting_column), (ending_row, ending_column))::_ -> (
+            Log.error "Entry for `%s` found near lines %d to %d column %d to %d" element starting_row ending_row starting_column ending_column)
+          | [] | _ -> ()
+          in
+          print_locations rest
+        )
+        in
+        print_locations positions;
+        failwith (Format.sprintf "Duplicate entries are not allowed.");
```
onionymous, line 565:
```suggestion
        Log.error "Duplicate entry for %s `%s` found:" kind element;
        let section = kind ^ "s" in
        let positions = List.map configurations ~f:(get_location section element) in
        let rec print_locations locations =
          match locations with
          | [] -> ()
          | location_list :: rest ->
              (match location_list with
              | [((starting_row, starting_column), (ending_row, ending_column))]
              | ((starting_row, starting_column), (ending_row, ending_column)) :: _ ->
                  Log.error
                    "Entry for `%s` found near lines %d to %d column %d to %d"
                    element
                    starting_row
                    ending_row
                    starting_column
                    ending_column
              | _ -> ());
              print_locations rest
        in
        print_locations positions;
        failwith (Format.sprintf "Duplicate entries are not allowed."));
```

onionymous (commented):
Just some minor syntax errors :) The entire `match` statement on line 552 is of `unit` type, but that doesn't mean we can ignore its value. Rather than:
```
match foo with
  | _ -> ()
in
baz
```
either discard the value and write it as a `let` expression:
```
let _ = match foo with
  | _ -> ()
in
baz
```

or use the semicolon to sequence statements:
```
( match foo with
    | _ -> ()
);
baz;
```

and the same thing on line line 549 with the `Log`, I think. (`Log` returns unit type as well)

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Ah @onionymous thank you for taking a look. It makes sense!

abishekvashok:
Making slow but definite progress :) @onionymous 
Now the below line seems to cause
```Ocaml
validate configuration (List.map ~f:(Json.Util.to_string) configurations);
```
this error:
```
Pyre encountered an internal exception: ("Yojson.Safe.Util.Type_error(\"Expected string, got object\", _)")
```

Is this in some way related because configurations is piped through `Option.some` ?

onionymous:
> Making slow but definite progress :) @onionymous Now the below line seems to cause
> 
> ```ocaml
> validate configuration (List.map ~f:(Json.Util.to_string) configurations);
> ```
> 
> this error:
> 
> ```
> Pyre encountered an internal exception: ("Yojson.Safe.Util.Type_error(\"Expected string, got object\", _)")
> ```
> 
> Is this in some way related because configurations is piped through `Option.some` ?

think we resolved this in OH last week - let me know if you have more issues! :)

abishekvashok:
@onionymous thank you. Yes we resolved it in the last OH. I am now a facing a bit of trouble in the logic of the inner `begin` keywords. But I will try to resolve it asap and will reach out if I get stuck :)

abishekvashok:
@onionymous almost done! Now I am getting the only the first occurrence of the duplicate entry.

abishekvashok:
@onionymous @0xedward thanks for all the help in this PR :)

0xedward (changes requested):
Yay! It looks like it's working now :) Would you mind adding some tests?

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -523,13 +522,49 @@ let parse source_jsons =
   }
 
 
-let validate { sources; sinks; features; _ } =
+let get_location ?encoding section label src =
```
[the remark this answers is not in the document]
arthaud: I believe this function always returns an empty list? `loop` is always called with `position` unchanged, and initialized to `[]`.
Also, it looks like `Jsonm.decode` only does parsing into tokens. What happens if a label or section appears twice in the json? We would probably error on the wrong line?

For instance:
```
 [
  {
    "source": ... # error here
  },
  {
    "source": ...
    "source": ... # duplicate here
  }
]
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -523,13 +522,49 @@ let parse source_jsons =
...
   let ensure_list_unique ~kind ~get_name elements =
     let seen = String.Hash_set.create () in
     let ensure_unique element =
       let element = get_name element in
-      if Hash_set.mem seen element then
-        failwith (Format.sprintf "Duplicate entry for %s: `%s`" kind element);
+      if Hash_set.mem seen element then (
+        Log.error "Duplicate entry for %s `%s` found:" kind element;
+        let section = kind ^ "s" in
+        let _ = List.map configurations ~f:(get_location section element) in
```
arthaud, line 565: We should make `get_location` side-effect free, return the locations and print them here.

arthaud (changes requested):

abishekvashok:
@arthaud thanks for looking into this. I took a short break - sorry for the late reply.

> Also, it looks like Jsonm.decode only does parsing into tokens. What happens if a label or section appears twice in the json? We would probably error on the wrong line?

That's why I implemented a counter inside which keeps track of preceding tokens. But I think it's hacky, and I trying to find a better way to do it and update this PR :)

abishekvashok:
Closing in favour of https://github.com/facebook/pyre-check/pull/732

> Also, it looks like Jsonm.decode only does parsing into tokens. What happens if a label or section appears twice in the json? We would probably error on the wrong line?

Fixed this in the new PR by changing the approach we took

PR #603, opened 2022-04-18
lazy parsing: Make AstEnvironment lazy

stroxler:
Summary:
This commit makes AstEnvironment lazy.

We do this in a few steps:
- Rename `RawSources` to `ReadWriteRawSources`. I'm not thrilled
  about this name but it's only used internally, and I think I can
  refactor to something more reasonable later on.
- Create a new `RawSources` module that is a read-only lazy view of
  `RawSources` with just a `get` function.
- Force the read-only record in `AstEnvironment.read_only` rely
  exclusively on `RawSources.get` rather than `ReadWriteRawSources.get`.
- Change the update logic:
  - a cold start update should parse nothing, but invalidate all
    internal modules (type check relies on the invalidated modules
    list to know what to check).
  - both shallow and fine grained updates should
    - clear all modules that changed:
      - modules requiring reparse
      - new implicit modules, which might come from a deleted __init__
      - deleted modules, which need to be wiped
    - reparse all modules corresponding to changed files if those
      modules already exist in ReadWriteRawSources
  - updates should return as implicit modules both all modules
    that changed, plus all internal modules with changes (this
    is needed for new internal paths, otherwise we'll fail to
    run typecheck on them downstream), plus all modules whose
    preprocessing step registered a dependency on a changed
    raw source.

Differential Revision: D35659549

### DIFF
--- a/source/analysis/astEnvironment.ml
+++ b/source/analysis/astEnvironment.ml
@@ -35,7 +35,7 @@ module RawSourceValue = struct
   let compare = Result.compare Source.compare ParserError.compare
 end
 
-module RawSources = struct
+module ReadWriteRawSources = struct
   include
     DependencyTrackedMemory.DependencyTrackedTableNoCache
       (SharedMemoryKeys.ReferenceKey)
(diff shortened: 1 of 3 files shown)


stroxler:
Closing - the underlying internal commit has been merged

PR #618, opened 2022-06-11
Drop support for Python 3.6, update `setup.py`

stroxler:
Summary:
Python 3.6 hit end-of-life support a couple months ago.

As a result, it's okay for us to drop support for it (which we actually
broke about 6 weeks ago anyway when we started using some metadata from
the `ast` module that doesn't exist until 3.7!)

So, I'm removing it from the tox setup; I also realized our `setup.py`
is way out of date and claimed we support 3.5 - 3.6 but we currently support
3.7 - 3.9 (we probably support 3.10 but I'll add that in a separate
commit because, until we hook up tests, that's not technically known).

Differential Revision: D37091837


stroxler:
Closing - I just needed the PR to run CI

PR #619, opened 2022-06-11
Use a try/except to degrade gracefully in 3.7

stroxler:
Summary:
I had hoped that we could bypass failures in the `find_symbols` module by
dropping support for Python 3.6, which is past end-of-life, but it turns
out the API we're using doesn't exist in 3.7 either and that's still a year
away from end-of-life.

So, for now we should catch the resulting AttributeError and degrade
gracefully.

Differential Revision: D37091905


stroxler:
I needed this PR to test github CI, closing now

PR #620, opened 2022-06-11
Fix type errors in `application.py`

stroxler:
Summary:
The flask stubs are incomplete, and we actually had an incorrect
type (`jsonify` returns a `Response` object, not a `str`)

Differential Revision: D37091874


stroxler:
Closing, I needed this to test github CI

PR #621, opened 2022-06-11
What happens if we bound click?

stroxler:
Differential Revision: D37092169


stroxler:
This didn't do the trick

PR #623, opened 2022-06-11
Update the pysa json in the docs / integ test

stroxler:
Summary:
This test has been broken for months.

I think half the problem is a slight change in the output of the SQL injection
description, which is more readable now.

The other half is that there's a new error that wasn't part of the example.

Differential Revision: D37092209


stroxler:
Can I reopen?

PR #630, opened 2022-06-15
avoid importing `typing.TypedDict`

stroxler:
Summary:
We export `pyre_extensions.tests` in the `__init__.py`,
which is surprising but also intended. So any lack of
backward compatibility with python will break not only
the tests but the entire extensions module even in production.

Now that we test pyre_extensions in github CI, we should be
able to catch such problems prior to release in the future.

Differential Revision: D37178251


d4l3k:
@stroxler are you planning on cutting a hotfix release? This is affecting downstream projects like TorchX/Ax

https://github.com/facebook/Ax/runs/6891472698?check_suite_focus=true

https://github.com/facebook/Ax/pull/1001

Issue #659, opened 2022-09-10
`Self` type in generic class reports error on line "-1"


bigfootjon:
# Pyre Bug

**Bug description**
In a generic class attempting to use `Self` as the return type causes a pyre error on line "-1".

**Reproduction steps**
```py
# pyre-strict

from typing import Generic, Self, TypeVar

TFoo = TypeVar("TFoo")

class GenericType(Generic[TFoo]):
    def get_self(self) -> Self:
        return self
```

**Expected behavior**
There should be no errors in this file. The return type of `get_self` should be inferred as `GenericType[TFoo]`.

**Logs**
Pyre output:
```
demo.py:-1:-1 Invalid type parameters [24]: Generic type `GenericType` expects 1 type parameter.
```

Please run your reproduction steps followed by `pyre rage > pyre_rage.log`, and upload the file here:

[pyre_rage.log](https://github.com/facebook/pyre-check/files/9539302/pyre_rage.log)

**Additional context**
No third party deps are needed to repro

bigfootjon:
After reading the code a bit, it seems this is a known problem:

https://github.com/facebook/pyre-check/blob/main/source/analysis/test/integration/typeVariableTest.ml#L3759

bigfootjon:
After discussing with @pradeep90 it seems the only way to support this is to synthesize a pyre-ignore on the synthesized `TypeVar`. This can be added with `TypecheckFlags`

PR #678, opened 2022-10-12
Support generic Self type

bigfootjon:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Basically, enable this example code to work:

```python3
# pyre-strict

from typing import Generic, Self, TypeVar

T = TypeVar("T")

class Foo(Generic[T]):
    def __init__(self, t: T) -> None:
        pass

    def get_self(self) -> Self:
        return Self
```

Before this change, this returns:

`Invalid type parameters [24]: Generic type Foo expects 1 type parameter.`

After, there are no errors reported.

## Test Plan

Ran the test suite, including the updated tests. I also ran on a private project that uses generic Self and no errors were reported.

Fixes: https://github.com/facebook/pyre-check/issues/659

### DIFF
--- a/source/analysis/postprocessing.ml
+++ b/source/analysis/postprocessing.ml
@@ -17,11 +17,10 @@ module Error = AnalysisError
    by one, and remove the used codes from the map of unused ignores. Since the hash tables are
    initialized with only the sources we're considering, this is sufficient to determine all ignored
    errors and unused ignores. *)
-let handle_ignores_and_fixmes
-    ~qualifier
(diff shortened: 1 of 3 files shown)


bigfootjon:
I'm not sure what I'm doing wrong, but I can see that the `TypecheckFlags` are being updated in the pre-processing step, but by the post-processing step the `ignore_lines` are wiped out (see logging I added).

I've tried tracing the code around and can't figure out what's going on :(

PR #720, opened 2023-03-07
Adds support for checking python version in models

abishekvashok:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Adds support for checking python versions in pysa models by extending the model parser to parse if-conditions as well.

Previously all if-conditions in the modes file were discarded with a model verification error of type Unsupportedstatement. Extend the AST logic to check for the presence of if conditions and honour them. Currently, we can check against the python version in the system provided to the taintAnalysis Command and while running the server by accessing the configuration.

This allows us to write if conditions along with else if blocks if needed in model files of the form:
`if sys.version <operator> (<python_major_version>, <python_minor_version>, <python_micro_version>):`
The models in the if body or the else body will be parsed only if the condition evaluates to true. sys.version is the only allowed left hand expression, all other expressions lead to a model verification error.

Nesting of if conditions are also supported. All comparison operators except in and not in are supported. If the right side expression is not a tuple, the if condition is ignored and a model verification error is emitted.

Update existing tests and ModelParser.parse invocations to pass in the python_version. Adds new unit tests as well.

Also adds documentation for the same.

## Test Plan

- Build the project from source using dune
- cd a repository down
- For ease, modify the model in documentation/pysa_tutorials/exercise1 to add an if condition with a comparison operator to it, eg:
```
if sys.version < (2,1,0):
    django.http.request.HttpRequest.GET: TaintSource[CustomUserControlled] = ...
    def eval(__source: TaintSink[CodeExecution], __globals, __locals): ...
```
- Create the following .pyre_configuration file:
```
{
    "source_directories": ["pyre-check/documentation/pysa_tutorial/exercise1/"],
    "search_path": ["./stubs/stubs"],
    "taint_models_path": "./pyre-check/documentation/pysa_tutorial/exercise1/"
}
```
- run pysa and see that the model is not getting detected. (No issues):

Screenshot 2023-03-20 at 11 44 18 AM

- change the model such that the if condition evaluates to true
```
if sys.version > (2,1,0):
    django.http.request.HttpRequest.GET: TaintSource[CustomUserControlled] = ...
    def eval(__source: TaintSink[CodeExecution], __globals, __locals): ...
```
- run pysa again and see that the model does get detected (1 issue found):

Screenshot 2023-03-20 at 11 45 13 AM

- change the if condition to something else, for example, `if (2,1,0):` to see a model verification error:

Screenshot 2023-03-20 at 11 46 49 AM

- run pysa on delibrately_vulnerable_app to make sure this doesn't break anything (github action)
- unit tests and github actions

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/83

### DIFF
--- a/source/configuration.ml
+++ b/source/configuration.ml
@@ -233,7 +233,7 @@ module PythonVersion = struct
     minor: int;
     micro: int;
   }
-  [@@deriving sexp, compare, hash, yojson]
+  [@@deriving sexp, compare, hash, yojson, equal]
 
   let default =
     {
--- a/source/interprocedural_analyses/taint/modelParser.ml
+++ b/source/interprocedural_analyses/taint/modelParser.ml
@@ -23,6 +23,54 @@ open Statement
 open Domains
 open ModelParseResult
 
+module PythonVersion = struct
+  (* Not putting the functions there to prevent circular dependency errors *)
+  include Configuration.PythonVersion
+
+  let parse_from_tuple tuple =
+    let parse_element = function
+      | Some { Node.value = Ast.Expression.Expression.Constant const; _ } -> (
+          match const with
+          | Integer integer -> Some (Ok integer)
+          | _ -> Some (Error "Only integer values are allowed inside the version tuple."))
+      | Some _ -> Some (Error "The value in the version tuple must be an integer constant.")
+      | None -> None
+    in
+    let open Core.Result in
+    (* we need at least a single value to emulate python tuple comparison *)
+    Option.value ~default:(Error "The tuple must not be empty") (parse_element (List.nth tuple 0))
+    >>= fun major ->
+    Option.value ~default:(Ok 0) (parse_element (List.nth tuple 1))
+    >>= fun minor ->
+    Option.value ~default:(Ok 0) (parse_element (List.nth tuple 2))
+    >>| fun micro -> { major; minor; micro }
+
+
+  let from_configuration
+      {
+        Configuration.Analysis.python_major_version = major;
+        python_minor_version = minor;
+        python_micro_version = micro;
+        _;
+      }
+    =
+    { major; minor; micro }
+
+
+  let compare_with left operator right =
+    match operator with
+    | ComparisonOperator.Equals -> Ok (equal left right)
+    | NotEquals -> Ok (not (equal left right))
+    | GreaterThan -> Ok (compare left right > 0)
+    | GreaterThanOrEquals -> Ok (compare left right >= 0)
+    | LessThan -> Ok (compare left right < 0)
+    | LessThanOrEquals -> Ok (compare left right <= 0)
+    | In -> Error ComparisonOperator.In
+    | NotIn -> Error ComparisonOperator.NotIn
+    | Is -> Error ComparisonOperator.Is
+    | IsNot -> Error ComparisonOperator.IsNot
+end
+
 let model_verification_error ~path ~location kind = { ModelVerificationError.kind; path; location }
 
 (* We don't have real models for attributes, so we make a fake callable model with a 'parameter'
@@ -2850,6 +2898,7 @@ let rec parse_statement
     ~source_sink_filter
     ~callables
     ~stubs
+    ~python_version
     statement
   =
   let open Core.Result in
@@ -2879,9 +2928,12 @@ let rec parse_statement
         | Some _ -> Target.create_method name
         | None -> Target.create_function name
       in
-      Ok [ParsedSignature { signature; location; call_target }]
+      [Ok (ParsedSignature { signature; location; call_target })]
   | { Node.value = Statement.Define { signature = { name; _ }; _ }; location } ->
-      Error [model_verification_error ~path ~location (DefineBodyNotEllipsis (Reference.show name))]
+      [
+        Error
+          (model_verification_error ~path ~location (DefineBodyNotEllipsis (Reference.show name)));
+      ]
   | {
    Node.value =
      Class
@@ -3014,11 +3066,13 @@ let rec parse_statement
 
                List.concat_map body ~f:signature)
         |> Option.value ~default:[]
-        |> return
+        |> List.map ~f:return
       else
-        Ok []
+        []
   | { Node.value = Class { Class.name; _ }; location } ->
-      Error [model_verification_error ~path ~location (ClassBodyNotEllipsis (Reference.show name))]
+      [
+        Error (model_verification_error ~path ~location (ClassBodyNotEllipsis (Reference.show name)));
+      ]
   | {
    Node.value =
      Assign
@@ -3026,20 +3080,20 @@ let rec parse_statement
    location;
   } ->
       if not (is_simple_name name) then
-        Error [model_verification_error ~path ~location (InvalidIdentifier target)]
+        [Error (model_verification_error ~path ~location (InvalidIdentifier target))]
       else if Expression.show annotation |> String.is_prefix ~prefix:"Sanitize[TaintInTaintOut["
       then
-        Error
-          [
-            model_verification_error
-              ~path
-              ~location
-              (InvalidTaintAnnotation
-                 {
-                   taint_annotation = annotation;
-                   reason = "TaintInTaintOut sanitizers cannot be modelled on attributes";
-                 });
-          ]
+        [
+          Error
+            (model_verification_error
+               ~path
+               ~location
+               (InvalidTaintAnnotation
+                  {
+                    taint_annotation = annotation;
+                    reason = "TaintInTaintOut sanitizers cannot be modelled on attributes";
+                  }));
+        ]
       else if
         Expression.show annotation |> String.equal "Sanitize"
         || Expression.show annotation |> String.is_prefix ~prefix:"Sanitize[TaintSource"
@@ -3057,72 +3111,75 @@ let rec parse_statement
             arguments;
           }
         in
-        Ok
-          [
-            ParsedAttribute
-              {
-                name;
-                source_annotation = None;
-                sink_annotation = None;
-                decorators = [decorator];
-                location;
-                call_target = Target.create_object name;
-              };
-          ]
+        [
+          Ok
+            (ParsedAttribute
+               {
+                 name;
+                 source_annotation = None;
+                 sink_annotation = None;
+                 decorators = [decorator];
+                 location;
+                 call_target = Target.create_object name;
+               });
+        ]
       else if Expression.show annotation |> String.is_substring ~substring:"TaintSource[" then
         let name = name_to_reference_exn name in
-        Ok
-          [
-            ParsedAttribute
-              {
-                name;
-                source_annotation = Some annotation;
-                sink_annotation = None;
-                decorators = [];
-                location;
-                call_target = Target.create_object name;
-              };
-          ]
+        [
+          Ok
+            (ParsedAttribute
+               {
+                 name;
+                 source_annotation = Some annotation;
+                 sink_annotation = None;
+                 decorators = [];
+                 location;
+                 call_target = Target.create_object name;
+               });
+        ]
       else if
         Expression.show annotation |> String.is_substring ~substring:"TaintSink["
         || Expression.show annotation |> String.is_substring ~substring:"TaintInTaintOut["
       then
         let name = name_to_reference_exn name in
-        Ok
-          [
-            ParsedAttribute
-              {
-                name;
-                source_annotation = None;
-                sink_annotation = Some annotation;
-                decorators = [];
-                location;
-                call_target = Target.create_object name;
-              };
-          ]
+        [
+          Ok
+            (ParsedAttribute
+               {
+                 name;
+                 source_annotation = None;
+                 sink_annotation = Some annotation;
+                 decorators = [];
+                 location;
+                 call_target = Target.create_object name;
+               });
+        ]
       else if Expression.show annotation |> String.equal "ViaTypeOf" then
         let name = name_to_reference_exn name in
-        Ok
-          [
-            ParsedAttribute
-              {
-                name;
-                source_annotation = None;
-                sink_annotation = Some annotation;
-                decorators = [];
-                location;
-                call_target = Target.create_object name;
-              };
-          ]
+        [
+          Ok
+            (ParsedAttribute
+               {
+                 name;
+                 source_annotation = None;
+                 sink_annotation = Some annotation;
+                 decorators = [];
+                 location;
+                 call_target = Target.create_object name;
+               });
+        ]
       else
-        Error
-          [
-            model_verification_error
-              ~path
-              ~location
-              (InvalidTaintAnnotation
-                 { taint_annotation = annotation; reason = "Unsupported annotation for attributes" });
-          ]
+        [
+          Error
+            (model_verification_error
+               ~path
+               ~location
+               (InvalidTaintAnnotation
+                  {
+                    taint_annotation = annotation;
+                    reason = "Unsupported annotation for attributes";
+                  }));
+        ]
   | {
    Node.value =
      Expression
@@ -3136,7 +3193,7 @@ let rec parse_statement
          _;
        };
    location;
-  } ->
+  } -> (
       let parse_model_sources ~name ~path expression =
         match expression with
         | Some ({ Node.value; location } as expression) ->
@@ -3175,12 +3232,12 @@ let rec parse_statement
                        ~taint_configuration
                        ~source_sink_filter
                        ~callables
-                       ~stubs)
+                       ~stubs
+                       ~python_version)
+            >>| List.concat
             >>| List.partition_result
             >>| (fun (results, errors) ->
-                  let results, remainder =
-                    List.zip_with_remainder (List.concat results) model_strings
-                  in
+                  let results, remainder = List.zip_with_remainder results model_strings in
                   let parsed_results, parsed_errors =
                     List.partition_map results ~f:(fun (result, model_source) ->
                         match result with
@@ -3203,7 +3260,7 @@ let rec parse_statement
                                       models_clause = Option.value_exn expression;
                                     })))
                   in
-                  let errors = List.concat errors @ parsed_errors in
+                  let errors = errors @ parsed_errors in
                   if List.is_empty errors then
                     match remainder with
                     | None -> Ok parsed_results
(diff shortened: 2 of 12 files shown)


source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -129,6 +129,9 @@ let parse_models_and_queries_from_sources
     ~source_sink_filter
     ~callables
     ~stubs
+    ~python_major_version
+    ~python_minor_version
+    ~python_micro_version
```
tianhan0, line 134: Could we create a module for this in `modelParser.ml` and `modelParser.mli` to be less verbose? Maybe something like

```
module PythonVersion = struct
  type t = {
    major_version: int;
    minor_version: int;
    micro_version: int;
  }
end
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3264,6 +3270,131 @@ let rec parse_statement
...
+              true
+            else if major_version == major then
+              if minor_version > minor then
+                true
+              else if minor_version == minor then
+                if micro_version >= micro then true else false
+              else
+                false
+            else
+              false
```
tianhan0, line 3355: I wonder if we can compute this by combing the logic for determining `Equals` and `GreaterThan`, since `GreaterThanOrEquals` is equivalent to `Equals` or `GreaterThan`?

source/interprocedural_analyses/taint/taintAnalysis.ml
tianhan0, line 134: To improve the readability, could we create a module for this in `modelParser.ml` and `modelParser.mli` and move your current parsing logic here? Maybe something like

```
module PythonVersion = struct
  type t = {
    major_version: int;
    minor_version: int;
    micro_version: int;
  }
  
  (* Move your current parsing logic in modelParser.ml from line 3273 to line 3394 here *)
end
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3264,6 +3270,131 @@ let rec parse_statement
...
+            | _ -> 0)
+          tuple
+      in
+      let major, minor, micro =
+        match version_list with
+        | [first; second; third; _] -> first, second, third
+        | [first; second; third] -> first, second, third
+        | [first; second] -> first, second, 0
+        | [first] -> first, 0, 0
+        | _ -> 0, 0, 0
```
tianhan0, line 3310: Can we create a function that parses a list of version numbers into a `PythonVersion.t`. Additionally, module `PythonVersion` can have a default version value of
```
{major_version = 0; minor_version = 0; micro_version = 0}
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3264,6 +3270,131 @@ let rec parse_statement
...
+      let major_version = extract_version python_major_version in
+      let minor_version = extract_version python_minor_version in
+      let micro_version = extract_version python_micro_version in
+      let evaluation_result =
+        match operator with
+        | Equals ->
+            if major_version == major && minor_version == minor && micro_version == micro then
+              true
+            else
+              false
```
tianhan0, line 3326: This can be simplified as
```
major_version == major && minor_version == minor && micro_version == micro
```

ditto below for `NotEquals`

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3264,6 +3270,131 @@ let rec parse_statement
...
+              true
+            else if major_version == major then
+              if minor_version < minor then
+                true
+              else if minor_version == minor then
+                if minor_version < minor then true else false
+              else
+                false
+            else
+              false
```
tianhan0, line 3367: Can this be the negation of `GreaterThanOrEquals`?

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3264,6 +3270,131 @@ let rec parse_statement
...
+        | GreaterThan ->
+            if major_version > major then
+              true
+            else if major_version == major then
+              if minor_version > minor then
+                true
+              else if minor_version == minor then
+                if micro_version > micro then true else false
```
tianhan0, line 3339: This can be simplified as `micro_version > micro`

tianhan0 (changes requested):

tianhan0:
There is also a build error that appears relevant? https://github.com/facebook/pyre-check/actions/runs/4354699634/jobs/7610329475

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: I tried to do so :)

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: That's a nice suggestion

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: Yes!

abishekvashok:
@tianhan0 about the build error, I tried list concatenation and it didn't work -> would appreciate some inputs there.
I tried to refractor the code to your suggestions -> but my Ocaml background is very much limited.

I am currently facing this weird error in which I cannot match python_version variable with the PythonVersion Constructor!

```
3333 |         | PythonVersion version -> version
                 ^^^^^^^^^^^^^
Error: Unbound constructor PythonVersion
make: *** [dev] Error 1  
```

tianhan0 (changes requested):

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3290,96 +3318,35 @@ let rec parse_statement
...
-            if major_version == major && minor_version == minor && micro_version == micro then
-              true
-            else
-              false
-        | NotEquals ->
-            if major_version != major || minor_version != minor || micro_version != micro then
-              true
-            else
-              false
+        | ComparisonOperator.Equals -> system_version = test_version
```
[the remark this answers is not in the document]
abishekvashok: I've used `.equal`, I think it's the same result?!

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3290,96 +3318,35 @@ let rec parse_statement
...
+              if system_version.minor > test_version.minor then
                 true
-              else if minor_version == minor then
-                if micro_version > micro then true else false
+              else if system_version.minor == test_version.minor then
+                system_version.micro > micro
               else
                 false
             else
               false
```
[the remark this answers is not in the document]
abishekvashok: Yep

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,38 @@ open Statement
...
 open ModelParseResult
 
+module PythonVersion = struct
+  type t = {
+    major: int;
+    minor: int;
+    micro: int;
+  }
+
+  let parse_from_integers major minor micro = { major; minor; micro }
```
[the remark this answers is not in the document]
abishekvashok: Yep that solves the undefined constructor problem as well -thanks!

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,38 @@ open Statement
...
+    in
+    let major, minor, micro =
+      match elements expression with
+      | [first; second; third; _] -> first, second, third
+      | [first; second; third] -> first, second, third
+      | [first; second] -> first, second, 0
+      | [first] -> first, 0, 0
+      | _ -> 0, 0, 0
+    in
+    { major; minor; micro }
```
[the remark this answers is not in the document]
abishekvashok: Yep!

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3290,96 +3318,35 @@ let rec parse_statement
...
+      let system_version =
+        match python_version with
+        | PythonVersion version -> version
+        | None -> PythonVersion.get_empty
       in
-      let major_version = extract_version python_major_version in
-      let minor_version = extract_version python_minor_version in
-      let micro_version = extract_version python_micro_version in
-      let evaluation_result =
+      let rec evaluate operator test_version system_version =
```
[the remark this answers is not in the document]
abishekvashok: Sure :)

abishekvashok:
@tianhan0 thanks for the wonderful suggestions - I've reworked the PR :)

For the build error, think about it further, it raises the questions for me -> since there can be a block of statements in the body of an if-block, wouldn't an error result in any of the expressions cancel all of them out? Return type for parse_statements is `(parsed_statements list, ModelVerficationError.t list) result`, that is either only a list of errors or a list of parsed statements. I am trying to think of a way to not change the return type and have a similar result...

abishekvashok:
Okay I solved it by converting the results of parse_statement to be a result list rather than a result object.
Please don't merge it just yet, I am trying to bundle in some test cases so that we can have coverage as well :)

tianhan0:
@abishekvashok I will defer the future reviews of this pull request to @arthaud, since he probably has more context on this. Thanks for making the changes based on my comments!

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3220,52 +3299,96 @@ let rec parse_statement
...
+       {
+         If.body;
+         If.test =
+           {
+             Node.value =
+               ComparisonOperator
+                 {
+                   ComparisonOperator.right = { Node.value = Tuple tuple; _ };
+                   ComparisonOperator.operator;
+                   _;
```
arthaud, line 3361: This means we silently assume the left hand side is `sys.version`? Not a fan of this.
That would mean we accept things like `if "hello" < (3, 6, 0)`
Could we only accept `sys.version`? It should be an `Identifier` AST node.

arthaud (changes requested):
Thanks for the contribution!
Could we also document this?

source/interprocedural_analyses/taint/modelParser.mli
```diff
@@ -5,39 +5,58 @@
...
+  python_version:PythonVersion.t ->
   unit ->
   ModelParseResult.t
 
 val verify_model_syntax : path:PyrePath.t -> source:string -> unit
 
-val parse_access_path
-  :  path:PyrePath.t option ->
+val parse_access_path :
+   path:PyrePath.t option ->
```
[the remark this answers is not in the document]
abishekvashok: @arthuad i am using `ocamlformat`. The only arguments I pass are the filename and `-i` for formatting inplace. I think arc uses ocamlformat as well but i don't remember the options. Can you check if any parameters are used internally?

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,61 @@ open Statement
 open Domains
 open ModelParseResult
 
+module PythonVersion = struct
```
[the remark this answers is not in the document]
abishekvashok: That's a good suggestion :)

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,61 @@ open Statement
...
+
+  let rec compare left operator right =
+    match operator with
+    | ComparisonOperator.Equals -> equal left right
+    | NotEquals -> not (compare left Equals right)
+    | GreaterThan -> greater_than left right
+    | GreaterThanOrEquals -> compare left Equals right || compare left GreaterThan right
+    | LessThan -> not (compare left GreaterThanOrEquals right)
+    | LessThanOrEquals -> not (compare left GreaterThan right)
+    | _ -> false
```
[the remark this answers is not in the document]
abishekvashok: In and NotIn, we could return a model verification error here ig

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -2861,9 +2917,12 @@ let rec parse_statement
         | Some _ -> Target.create_method name
         | None -> Target.create_function name
       in
-      Ok [ParsedSignature { signature; location; call_target }]
+      [Ok [ParsedSignature { signature; location; call_target }]]
```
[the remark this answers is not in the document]
abishekvashok: Previously,
`parse_statements` returned `(parsed_statements list, error list) result`
ie, either a list of parsed statements or errors, not both.

The way we have the if statement is that there would be many more statements bundled inside the body, the problem therefore if we don't change parsed_statement into a result list is that if there is an error in the if-body, we would not be able to return the non-errored, parsed statements.

So, if we just return result, we pretty much make --no-verify for if based models to not work.

Also, semantically i guess this makes sense as a statement (which can be a group of statements now) can return a list of results.

This doesn't produce any changes in the callsites of parse_statements as they work on a result list later on.

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: Okay, we can do that!

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,61 @@ open Statement
...
+      | _ -> default_version
+    in
+    let versions = List.map ~f:from_element tuple in
+    match versions with
+    | [major; minor; micro; _]
+    | [major; minor; micro] ->
+        { major; minor; micro }
+    | [major; minor] -> { major; minor; micro = default_version }
+    | [major] -> { major; minor = default_version; micro = default_version }
+    | _ -> empty
```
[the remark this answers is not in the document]
arthaud: Yes we can have empty tuples. We could also have tuples with more than 5 elements here, right?

source/interprocedural_analyses/taint/modelParser.ml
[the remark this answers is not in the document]
abishekvashok: @arthaud are you sure this works? ModelParser.PythonVersion.major then is unbound.
NB: I also modified the mli file so that `t = Configuration.PythonVersion.t`

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: @arthaud also I checked the return value of the real sys.version, which is type string.. Just a heads up that we are deviating a bit from the expected behavior and could use something like `python_version`

abishekvashok:
@arthaud made the requested changes and updated the PR body. Will push changes to the documentation now.

abishekvashok:
Here's the documentation changes as well. I can, for ease of review, move them to another PR to be stacked on top of this one in the internal tooling.

Screenshot 2023-03-20 at 12 15 51 PM

Thanks for the review and the suggestions!

source/configuration.ml
```diff
@@ -233,7 +233,7 @@ module PythonVersion = struct
     minor: int;
     micro: int;
   }
-  [@@deriving sexp, compare, hash, yojson]
+  [@@deriving sexp, compare, hash, yojson, equal, compare]
```
arthaud, line 236: Here and other places: `compare` was already present.

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,60 @@ open Statement
...
+    let micro_version = Option.value ~default:(Ok 0) (parse_element (List.nth tuple 2)) in
+    match major_version with
+    | Ok major -> (
+        match minor_version with
+        | Ok minor -> (
+            match micro_version with
+            | Ok micro -> Ok { major; minor; micro }
+            | Error error -> Error error)
+        | Error error -> Error error)
+    | Error error -> Error error
```
arthaud, line 53: Looks good! This could be simplified a bit though, using `Core.Result`:
```suggestion
    let open Core.Result in
    Option.value ~default:(Error "The tuple must not be empty") (parse_element (List.nth tuple 0))
    >>= fun major_version ->
    Option.value ~default:(Ok 0) (parse_element (List.nth tuple 1))
    >>= fun minor_version ->
    Option.value ~default:(Ok 0) (parse_element (List.nth tuple 2))
    >>| fun micro_version ->
    { major; minor; micro }
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,60 @@ open Statement
...
+
+
+  let from_configuration configuration =
+    {
+      major = Configuration.Analysis.(configuration.python_major_version);
```
arthaud, line 58: nitpick: To avoid that ugly (that's subjective*) `Configuration.Analysis.(configuration.python_major_version)` you could use pattern matching

```suggestion
  let from_configuration { Configuration.Analysis.python_major_version; python_minor_version; python_micro_version; _ } =
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -2995,33 +3053,36 @@ let rec parse_statement
                in
 
                List.concat_map body ~f:signature)
-        |> Option.value ~default:[]
-        |> return
+        |> function
+        | Some result -> List.map ~f:(fun res -> Ok res) result
+        | None -> []
```
arthaud, line 3058:
```suggestion
        |> Option.bind ~f:(List.map ~f:return)
        |> Option.value ~default:[]
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3220,52 +3284,122 @@ let rec parse_statement
...
+              location;
+              expected_models;
+              unexpected_models;
+            };
+        ]
+      in
+      match result with
+      | Ok parsed_statements_list ->
+          List.map ~f:(fun statement -> Ok statement) parsed_statements_list
+      | Error errors_list -> List.map ~f:(fun error -> Error error) errors_list)
```
arthaud, line 3336: Not a fan of the big `let result = ... in match result ..`. It's a bit hard to read since it forces me to backtrace.
I think we can do better.
First of, we can result directly `Ok (ParsedQuery ...)` line 3319.
For the error case, we might just need to update that `as_result_error_list` function.
If we need further transformation, you can always add a `Result.map_error` at the end.

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3220,52 +3284,122 @@ let rec parse_statement
...
+                       Node.value =
+                         Name
+                           (Name.Attribute
+                             {
+                               base = { Node.value = Name (Name.Identifier "sys"); _ };
+                               attribute = "version";
+                               _;
+                             });
+                       _;
+                     };
```
arthaud, line 3360: nit: move left before the right :p

source/interprocedural_analyses/taint/modelVerificationError.ml
```diff
@@ -404,6 +407,19 @@ let description error =
...
+         operator version_tuple`. All models inside the if-block (along with those in else-if and \
+         else block, if present) will be ignored."
+        (Expression.show condition)
+  | UnsupportedVersionConstant error ->
+      Format.sprintf "Unsupported element type in version tuple in if condition: %s" error
+  | UnsupportedComparisonOperator operator ->
+      let pp_operator =
+        Format.asprintf "%a" Expression.ComparisonOperator.pp_comparison_operator operator
+      in
+      Format.sprintf "The operator `%s` in the if condition is not supported" pp_operator
```
arthaud, line 422: You can use `Format.asprintf` with `%a` then
```suggestion
      Format.asprintf "The operator `%a` in the if condition is not supported" Expression.ComparisonOperator.pp_comparison_operator operator
```

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -165,13 +167,19 @@ let parse_models_and_queries_from_sources
...
+          { taint_model_paths; python_major_version; python_minor_version; python_micro_version; _ };
+        _;
+      }
     ~taint_configuration
     ~resolution
     ~source_sink_filter
     ~callables
     ~stubs
   =
+  let python_version = ModelParser.PythonVersion.from_integers python_major_version python_minor_version python_micro_version in
```
arthaud, line 182: Couldn't we use `Taint.ModelParser.PythonVersion.from_configuration` here too?

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,60 @@ open Statement
 open Domains
 open ModelParseResult
 
+module PythonVersion = struct
+  (* Not putting the functions there to prevent circular dependency errors *)
+  include Configuration.PythonVersion
+
+  let parse_from_tuple ~tuple =
```
arthaud, line 30: nitpick: since we only take a single parameter and it's pretty obvious from the name, I wouldn't use a named parameter
```suggestion
  let parse_from_tuple tuple =
```

arthaud (changes requested):

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: > (that's subjective*) 

I understand but I agree with the classification, the code you gave is much more beautiful ;)

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: @arthaud this gives type error for the `List.map` function used by Option.bind.
We could reverse back to the match clauses or use the following new code that I imagine will work:
```Ocaml
         |> Option.value ~default:[]
         |> List.map ~f:return
```

source/interprocedural_analyses/taint/taintAnalysis.ml
abishekvashok: Hm, yes if we don't pattern match above, I can do this :)

abishekvashok:
@arthuad sorry for the back and forth, but I fixed for most of the suggestions :)
The other two, I left comments. Let me know if they are satisfactory.

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -23,6 +23,54 @@ open Statement
...
+    let parse_element = function
+      | Some { Node.value = Ast.Expression.Expression.Constant const; _ } -> (
+          match const with
+          | Integer integer -> Some (Ok integer)
+          | _ -> Some (Error "Only integer values are allowed inside the version tuple."))
+      | Some _ -> Some (Error "The value in the version tuple must be an integer constant.")
+      | None -> None
+    in
+    let open Core.Result in
+    (* we need atleast a single value to emulate python tuple comparison *)
```
arthaud, line 40: nitpick: typo
```suggestion
    (* we need at least a single value to emulate python tuple comparison *)
```

source/interprocedural_analyses/taint/modelParser.ml
```diff
@@ -3220,52 +3277,122 @@ let rec parse_statement
...
+                 path;
+                 location;
+                 expected_models;
+                 unexpected_models;
+               });
+        ]
+      in
+      match parsed_statements_errors with
+      | Ok parsed_statements_list -> parsed_statements_list
+      | Error errors_list -> List.map ~f:(fun error -> Error error) errors_list)
```
arthaud, line 3329: To avoid the big `let parsed_statements_errors = ... in match parsed_statements_errors with ..` here could you do something like:
```suggestion
      |> function
      | Ok parsed_statements_list -> parsed_statements_list
      | Error errors_list -> List.map ~f:(fun error -> Error error) errors_list)
```

source/interprocedural_analyses/taint/test/testHelper.ml
```diff
@@ -428,6 +428,7 @@ let get_initial_models ~context =
       ~source_sink_filter:None
       ~callables:None
       ~stubs:(Target.HashSet.create ())
+      ~python_version:(ModelParser.PythonVersion.default)
```
arthaud, line 431: nit here and below: you don't need the parenthesis
```suggestion
      ~python_version:ModelParser.PythonVersion.default
```

arthaud (changes requested):
Looks pretty good, just a few nitpicks.

source/interprocedural_analyses/taint/modelParser.ml
abishekvashok: That's better!

abishekvashok:
@arthaud thanks for the review(s) and suggestion(s)! I've action your latest review :)

abishekvashok:
@arthaud you might want to reimport this - I downgraded ocamlformat in a new opam switch and some linter generated changes have disappeared. You might want to reimport :)

abishekvashok:
@arthaud done :)

arthaud (changes requested):
It looks like `make test` throws a few errors. Could you fix those?

```
File "interprocedural_analyses/taint/test/forwardAnalysisTest.ml", lines 42-49, characters 6-10:
42 | ......ModelParser.parse
43 |         ~resolution:global_resolution
44 |         ~source:models
45 |         ~taint_configuration:TaintConfiguration.Heap.default
46 |         ~source_sink_filter:None
47 |         ~callables:None
48 |         ~stubs:(Target.HashSet.create ())
49 |         ()
Warning 5 [ignored-partial-application]: this function application is partial,
maybe some arguments are missing.
File "interprocedural_analyses/taint/test/forwardAnalysisTest.ml", lines 42-49, characters 6-10:
42 | ......ModelParser.parse
43 |         ~resolution:global_resolution
44 |         ~source:models
45 |         ~taint_configuration:TaintConfiguration.Heap.default
46 |         ~source_sink_filter:None
47 |         ~callables:None
48 |         ~stubs:(Target.HashSet.create ())
49 |         ()
Error: This expression has type
         python_version:Taint.ModelParser.PythonVersion.t ->
         Taint__.ModelParseResult.t
       but an expression was expected of type Taint.ModelParseResult.t
File "interprocedural_analyses/taint/test/modelTest.ml", lines 97-104, characters 4-8:
 97 | ....ModelParser.parse
 98 |       ~resolution:global_resolution
 99 |       ~source
100 |       ~taint_configuration
101 |       ~source_sink_filter:(Some taint_configuration.source_sink_filter)
102 |       ~callables:None
103 |       ~stubs:(Target.HashSet.create ())
104 |       ()
Warning 5 [ignored-partial-application]: this function application is partial,
maybe some arguments are missing.
File "interprocedural_analyses/taint/test/modelTest.ml", lines 97-104, characters 4-8:
 97 | ....ModelParser.parse
 98 |       ~resolution:global_resolution
 99 |       ~source
100 |       ~taint_configuration
101 |       ~source_sink_filter:(Some taint_configuration.source_sink_filter)
102 |       ~callables:None
103 |       ~stubs:(Target.HashSet.create ())
104 |       ()
Error: This expression has type
         python_version:Taint.ModelParser.PythonVersion.t ->
         Taint__.ModelParseResult.t
       but an expression was expected of type Taint.ModelParseResult.t
............................................................................
```

Could you also add some unit tests? We have `modelTest.ml` to test the parser.

abishekvashok:
@arthaud fixed the existing tests and added some more. I do however get an error when I run `make test`:
```
File "dune", line 224, characters 3-44:   
224 |    pyrelib.internalTaintIntegrationTestFiles pyrelib.taintTestHelper))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Library "pyrelib.internalTaintIntegrationTestFiles" not found.
-> required by _build/default/decoratorsTest.exe
-> required by alias runtest in dune:218
...............................           
Ran: 31 tests in: 17.76 seconds.
OK
make: *** [test] Error 1
```

As far as I remember, this will not happen "internally", correct? Also, if so, can't we fix this?

arthaud:
Thanks for the contribution!

abishekvashok:
> Thanks for the contribution!

Always a pleasure ;)

PR #734, opened 2023-05-06
Implement json parser and migrate taintConfiguration

abishekvashok:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

## Summary

Implements a json parsing module inside jsonParsing.ml that adds support for line and column numbers based on Jsonm. Removes all dependencies of taint configuration parsing on YoJSON which doesn't have any support for positions.

Since, Jsonm is pretty bare bone, implemented some utility functions inside a Util module that does pretty much what YoJSON's utility functions did.

There are minor changes in taintConfiguration to accomodate for the change.

This is the base part of changes that seek to remove dependency completely on YoJSON and switch to JSONM because YoJSON doesn't support line numbers in any way which we require for descriptive error messages.

Modifies and updates confirmationTest. Since we use a proper json parser, the json files for testing have been updated to use the proper json format as well.

## Test Plan

- make
- make test ( only failing test is due to the internal taint integration files -> you know why (: )
<details>

<summary>See test results</summary>

```
../scripts/generate-version-number.sh: line 16: greadlink: command not found
Build info: Darwin x86_64 @ Thu May 11 2023 (development build)
../scripts/generate-version-number.sh: line 30: hg: command not found
Git commit: ced2d4cdac4d2010ff9d11ee5eec2ec21365046a
dune build @install -j auto --profile dev
File "dune", line 379, characters 2-43:   
379 |   pyrelib.internalTaintIntegrationTestFiles
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Library "pyrelib.internalTaintIntegrationTestFiles" not found.
-> required by _build/default/decoratorsTest.exe
-> required by alias runtest in dune:359
................                                    
Ran: 16 tests in: 0.67 seconds.
OK
............                                       
Ran: 12 tests in: 0.42 seconds.
OK
................                                   
Ran: 16 tests in: 0.38 seconds.
OK
.........................................          
Ran: 41 tests in: 0.57 seconds.
OK
.........................................          
Ran: 41 tests in: 0.74 seconds.
OK
...............................
Ran: 31 tests in: 0.45 seconds.
OK
................................                   
Ran: 32 tests in: 0.56 seconds.
OK
............                                       
Ran: 12 tests in: 0.32 seconds.
OK
...............................
[...]

PR #735, opened 2023-05-08
Display positions in RuleCodeDuplicate errors

abishekvashok:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Display positions of where duplicate rule codes appear in
RuleCodeDuplicate taint configuration errors.

Previously rule codes were only checked for uniqueness within a single
taint.config file. This leads to unintended results when multiple rules
share the same code in multiple taint.config files under analysis.
The errors emitted also lacked positioning information which makes it
harder to find where duplicate rule codes occurred.

Adds ability to cross validate rule codes as now like source and sink
uniqueness checks, rule codes are checked after all taint.config files
are parsed. Also Since we now store positioning information for all
parsed taint.config nodes, adds position information of the previous
and current location when a rule code appears.

Since uniqueness is checked after parsing, modifies Rules module to
contain the taint config path and location to be used in error
messages - created a module in jsonParsing.ml for the same.

Other minor changes:
- Update tests to check for the new sort of RuleDuplicate error
- Update tests to conform to the updated rule records

## Test Plan

- Before the changes, ran pysa on `documentaiton/pysa_tutorial/exercise1/`:

Screenshot 2023-05-06 at 1 17 36 PM

- After the changes with default `taint.config`:

Screenshot 2023-05-27 at 5 28 55 PM

- After the changes with the following `taint.config`:
```json
{
  "sources": [
    {
      "name": "CustomUserControlled",
      "comment": "use to annotate user input"
    }
  ],

  "sinks": [
    {
      "name": "CodeExecution",
      "comment": "use to annotate execution of python code"
    }
  ],

  "features": [],

  "rules": [
    {
      "name": "Possible RCE:",
      "code": 5001,
      "sources": [ "CustomUserControlled" ],
      "sinks": [ "CodeExecution" ],
      "message_format": "User specified data may reach a code execution sink"
    },
    {
      "name": "test-duplicate",
      "code": 5001,
      "sources": [ "CustomUserControlled" ],
      "sinks": [ "CodeExecution" ],
      "message_format": "duplicate"
    }
  ]
}
```

Screenshot 2023-05-31 at 2 06 36 PM

- Ran tests with `make test`.

Fixes part of https://github.com/MLH-Fellowship/pyre-check/issues/82

Footnotes:
- Pysa Github CI Action was failing before this PR

### DIFF
--- a/source/interprocedural_analyses/taint/rule.ml
+++ b/source/interprocedural_analyses/taint/rule.ml
@@ -18,6 +18,7 @@ type t = {
   code: int;
   name: string;
   message_format: string; (* format *)
+  location: JsonParsing.JsonAst.LocationWithPath.t option; (* Location where the rule was defined *)
 }
 [@@deriving compare, show]
 
--- a/source/interprocedural_analyses/taint/taintConfiguration.ml
+++ b/source/interprocedural_analyses/taint/taintConfiguration.ml
@@ -520,6 +520,7 @@ module Heap = struct
           name = "Possible shell injection.";
           message_format =
             "Possible remote code execution due to [{$sources}] data reaching [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "Test"; Sources.NamedSource "UserControlled"];
@@ -528,6 +529,7 @@ module Heap = struct
           code = 5002;
           name = "Test flow.";
           message_format = "Data from [{$sources}] source(s) may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "UserControlled"];
@@ -536,6 +538,7 @@ module Heap = struct
           code = 5005;
           name = "User controlled data to SQL execution.";
           message_format = "Data from [{$sources}] source(s) may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources =
@@ -547,6 +550,7 @@ module Heap = struct
           code = 5006;
           name = "Restricted data being logged.";
           message_format = "Data from [{$sources}] source(s) may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "UserControlled"];
@@ -555,6 +559,7 @@ module Heap = struct
           code = 5007;
           name = "User data to XML Parser.";
           message_format = "Data from [{$sources}] source(s) may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "UserControlled"];
@@ -563,6 +568,7 @@ module Heap = struct
           code = 5008;
           name = "XSS";
           message_format = "Possible XSS due to [{$sources}] data reaching [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "Demo"];
@@ -571,6 +577,7 @@ module Heap = struct
           code = 5009;
           name = "Demo flow.";
           message_format = "Data from [{$sources}] source(s) may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "UserControlled"];
@@ -579,6 +586,7 @@ module Heap = struct
           code = 5010;
           name = "User data to getattr.";
           message_format = "Attacker may control at least one argument to getattr(,).";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "Test"];
@@ -588,6 +596,7 @@ module Heap = struct
           name = "Flow with one transform.";
           message_format =
             "Data from [{$sources}] source(s) via [{$transforms}] may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "Test"];
@@ -597,6 +606,7 @@ module Heap = struct
           name = "Flow with two transforms.";
           message_format =
             "Data from [{$sources}] source(s) via [{$transforms}] may reach [{$sinks}] sink(s)";
+          location = None;
         };
         {
           sources = [Sources.NamedSource "Demo"];
@@ -606,6 +616,7 @@ module Heap = struct
           name = "Duplicate demo flow.";
           message_format =
             "Possible remote code execution due to [{$sources}] data reaching [{$sinks}] sink(s)";
+          location = None;
         };
       ]
     in
@@ -717,20 +728,23 @@ module Error = struct
         labels: string list;
       }
     | InvalidMultiSink of string
-    | RuleCodeDuplicate of int
+    | RuleCodeDuplicate of {
+        code: int;
+        previous_location: JsonAst.LocationWithPath.t option;
+      }
     | OptionDuplicate of string
     | SourceDuplicate of string
     | SinkDuplicate of string
     | TransformDuplicate of string
     | FeatureDuplicate of string
-  [@@deriving equal]
+  [@@deriving equal, compare]
 
   type t = {
     kind: kind;
     path: PyrePath.t option;
     location: JsonAst.Location.t option;
   }
-  [@@deriving equal]
+  [@@deriving equal, compare]
 
   let create_with_location ~path ~kind ~location =
     { kind; path = Some path; location = Some location }
@@ -780,8 +794,17 @@ module Error = struct
           sink
           (String.concat labels ~sep:", ")
     | InvalidMultiSink sink -> Format.fprintf formatter "`%s` is not a multi sink" sink
-    | RuleCodeDuplicate code ->
-        Format.fprintf formatter "Multiple rules share the same code `%d`" code
+    | RuleCodeDuplicate { code; previous_location = None } ->
+        Format.fprintf formatter "Multuple rules share the same code `%d`" code
+    | RuleCodeDuplicate { code; previous_location = Some previous_location } ->
+        Format.fprintf
+          formatter
+          "Multiple rules share the same code `%d`, previous rule was at `%a:%a`"
+          code
+          PyrePath.pp
+          previous_location.path
+          JsonAst.Location.pp_start
+          previous_location.location
     | OptionDuplicate name ->
         Format.fprintf formatter "Multiple values were passed in for option `%s`" name
     | SourceDuplicate name -> Format.fprintf formatter "Duplicate entry for source: `%s`" name
@@ -876,9 +899,12 @@ let from_json_list source_json_list =
     json_exception_to_error ~path ~section:key (fun () ->
         JsonAst.Json.Util.member_exn key value |> JsonAst.Json.Util.to_string_exn |> Result.return)
   in
-  let json_integer_member ~path key value =
+  let json_integer_member_with_location ~path key value =
     json_exception_to_error ~path ~section:key (fun () ->
-        JsonAst.Json.Util.member_exn key value |> JsonAst.Json.Util.to_int_exn |> Result.return)
+        let node = JsonAst.Json.Util.member_exn key value in
+        let value = JsonAst.Json.Util.to_int_exn node in
+        let location = JsonAst.Json.Util.to_location_exn node in
+        Result.return (value, location))
   in
   let array_member ~path ?section name json =
     let node = JsonAst.Json.Util.member name json in
@@ -976,14 +1002,6 @@ let from_json_list source_json_list =
     |> Result.combine_errors
     |> Result.map_error ~f:List.concat
   in
-  let seen_rules = Int.Hash_set.create () in
-  let validate_code_uniqueness ~path code =
-    if Hash_set.mem seen_rules code then
-      Result.Error [Error.create ~path ~kind:(Error.RuleCodeDuplicate code)]
-    else (
-      Hash_set.add seen_rules code;
-      Result.Ok ())
-  in
   let parse_source_reference ~path ~allowed_sources source =
     AnnotationParser.parse_source ~allowed:allowed_sources source
     |> Result.map_error ~f:(fun _ -> Error.create ~path ~kind:(Error.UnsupportedSource source))
@@ -1003,6 +1021,7 @@ let from_json_list source_json_list =
       name: string;
       message_format: string;
       code: int;
+      location: JsonAst.LocationWithPath.t;
     }
 
     let parse ~path json =
@@ -1010,9 +1029,9 @@ let from_json_list source_json_list =
       >>= fun name ->
       json_string_member ~path "message_format" json
       >>= fun message_format ->
-      json_integer_member ~path "code" json
-      >>= fun code ->
-      validate_code_uniqueness ~path code >>= fun () -> Result.Ok { name; message_format; code }
+      json_integer_member_with_location ~path "code" json
+      >>| fun (code, location) ->
+      { name; message_format; code; location = JsonAst.LocationWithPath.create ~path ~location }
   end
   in
   let parse_rules ~allowed_sources ~allowed_sinks ~allowed_transforms (path, json) =
@@ -1046,8 +1065,8 @@ let from_json_list source_json_list =
       |> Result.combine_errors
       >>= fun transforms ->
       RuleCommonAttributes.parse ~path json
-      >>= fun { name; message_format; code } ->
-      Result.Ok { Rule.sources; sinks; transforms; name; code; message_format }
+      >>| fun { name; message_format; code; location } ->
+      { Rule.sources; sinks; transforms; name; code; message_format; location = Some location }
     in
     array_member ~path "rules" json
     >>= fun rules ->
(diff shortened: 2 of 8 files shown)


arthaud (changes requested):
This depends on #734 so let's wait for it to go through first.

source/interprocedural_analyses/taint/rule.ml
```diff
@@ -18,6 +18,9 @@ type t = {
   code: int;
   name: string;
   message_format: string; (* format *)
+  (* taint config location information if exists*)
+  taint_config_path: PyrePath.t option;
+  taint_config_location: JsonParsing.Location.t option;
```
arthaud, line 23: Instead of storing it this way, what you could do is have a `Json.Location.WithPath.t` or `Json.LocationWithPath.t`
Also, why can it be `None`?

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -687,11 +709,41 @@ module SharedMemory = struct
       None
 end
 
+module TaintConfigLocation = struct
```
arthaud, line 712: ditto previous comment

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -687,11 +709,41 @@ module SharedMemory = struct
...
+end
+
 module Error = struct
   type kind =
     | FileNotFound
     | FileRead
-    | InvalidJson of string
+    | InvalidJson of {
+        message: string;
+        location: JsonParsing.Location.t;
```
arthaud, line 745: ditto previous PR, could we store the location directly in `type t`?

arthaud (changes requested):

source/interprocedural_analyses/taint/rule.ml
abishekvashok: Yes, `LocationWithPath.t` is better.

> Also, why can it be None?

We define some rules implicitly in the code

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Yes!

abishekvashok:
We'll wait for the other PR to go through and I will work on the suggestions for this one which will also collapse the diff a bit.

abishekvashok:
@arthaud @alexkassil since #734 is merged now, I kinda rewrote this PR :)

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -717,20 +728,23 @@ module Error = struct
...
+        code: int;
+        previous_location: JsonAst.LocationWithPath.t option;
+      }
     | OptionDuplicate of string
     | SourceDuplicate of string
     | SinkDuplicate of string
     | TransformDuplicate of string
     | FeatureDuplicate of string
-  [@@deriving equal]
+  [@@deriving equal, compare]
```
arthaud, line 740: Why do we need `compare` here (and below)?
```suggestion
  [@@deriving equal]
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -780,8 +794,12 @@ module Error = struct
           sink
           (String.concat labels ~sep:", ")
     | InvalidMultiSink sink -> Format.fprintf formatter "`%s` is not a multi sink" sink
-    | RuleCodeDuplicate code ->
-        Format.fprintf formatter "Multiple rules share the same code `%d`" code
+    | RuleCodeDuplicate { code; previous_location } ->
+        Format.fprintf
+          formatter
+          "Multiple rules share the same code `%d`. Previously seen in %s."
```
arthaud, line 800: nitpick:
```suggestion
          "Multiple rules share the same code `%d`, previous rule was at %s."
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1517,6 +1549,57 @@ let validate ({ Heap.sources; sinks; transforms; features; _ } as configuration)
...
+      | None ->
+          let _ = Hashtbl.add table ~key ~data:entry in
+          Result.Ok ()
+      | Some previous_entry -> (
+          if
+            (* Some generated rules can have the same code, compare the base defintions. *)
+            HashTableEntry.equal previous_entry entry
+          then
+            Result.Ok ()
+          else
```
arthaud, line 1580:
```suggestion
      | Some previous_entry when HashTableEntry.equal previous_entry entry -> 
            (* Some generated rules can have the same code, compare the base defintions. *)
            Result.Ok ()
      | Some previous_entry ->
```

arthaud (changes requested):

source/interprocedural_analyses/taint/rule.ml
```diff
@@ -18,6 +18,7 @@ type t = {
   code: int;
   name: string;
   message_format: string; (* format *)
+  location_with_path: JsonParsing.JsonAst.LocationWithPath.t option; (* Location where the rule was defined *)
```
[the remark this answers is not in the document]
abishekvashok: We can call it location.

Not sure if we make it non-optional what to do for implicitly defined rules. For example, consider these defined in taintConfiguration.ml, which position and path should we use?
https://github.com/facebook/pyre-check/blob/8c639dca3a937026cf6d14130964cc485923c066/source/interprocedural_analyses/taint/taintConfiguration.ml#L515-L610

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -780,8 +794,12 @@ module Error = struct
           sink
           (String.concat labels ~sep:", ")
     | InvalidMultiSink sink -> Format.fprintf formatter "`%s` is not a multi sink" sink
-    | RuleCodeDuplicate code ->
-        Format.fprintf formatter "Multiple rules share the same code `%d`" code
+    | RuleCodeDuplicate { code; previous_location } ->
```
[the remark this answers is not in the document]
abishekvashok: So the only place where previous_location is `None` is for rules defined implicitly in the binary. Will change the message accordingly and use two cases!

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -780,8 +794,12 @@ module Error = struct
...
+        Format.fprintf
+          formatter
+          "Multiple rules share the same code `%d`. Previously seen in %s."
+          code
+          (JsonAst.LocationWithPath.optional_to_string previous_location)
```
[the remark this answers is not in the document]
abishekvashok: So we'll just print the path, start line, and column for now. The end goal being the vs code extension code use this :)

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1003,16 +1016,23 @@ let from_json_list source_json_list =
       name: string;
       message_format: string;
       code: int;
+      location_with_path: JsonAst.LocationWithPath.t;
```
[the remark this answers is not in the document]
abishekvashok: Sure!

source/jsonParsing.ml
```diff
@@ -100,12 +106,28 @@ module JsonAst = struct
...
+    }
+    [@@deriving equal, show, compare]
+
+    let create ~location ~path = { location; path }
+
+    let optional_to_string optional_location_with_path =
+      match optional_location_with_path with
+      | Some { path; location } ->
+          Format.asprintf "`%s` %s" (PyrePath.show path) (Location.to_string location)
+      | None -> "unknown or internal location"
```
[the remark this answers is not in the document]
abishekvashok: Was going to use this in subsequent PRs but okay if we just print the start line and column :)

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1003,16 +1016,23 @@ let from_json_list source_json_list =
...
       >>= fun name ->
       json_string_member ~path "message_format" json
       >>= fun message_format ->
-      json_integer_member ~path "code" json
-      >>= fun code ->
-      validate_code_uniqueness ~path code >>= fun () -> Result.Ok { name; message_format; code }
+      json_integer_member_with_location ~path "code" json
+      >>= fun (code, location) ->
+      Result.Ok
+        {
```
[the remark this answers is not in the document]
abishekvashok: Perfect!

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1517,6 +1549,57 @@ let validate ({ Heap.sources; sinks; transforms; features; _ } as configuration)
...
+      type t =
+        | InternalLocation
+        | LocationWithPath of JsonAst.LocationWithPath.t
+      [@@deriving equal]
+
+      let create location_with_path =
+        match location_with_path with
+        | Some location_with_path -> LocationWithPath location_with_path
+        | None -> InternalLocation
+    end
```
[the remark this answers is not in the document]
abishekvashok: Okay!

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: So previously I used a List.`dedup and sort` to prevent raising two errors when a combined_source_rule and rule both contained the same code which required compare. Now, I migrated to use a Hashset which is more optimal for this comparison. We will need the comparison since a combined_source_rule can generate more than one rule..

source/interprocedural_analyses/taint/rule.ml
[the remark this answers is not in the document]
arthaud: I see. Alright, let's keep it optional then. Thanks for checking.

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -780,8 +794,21 @@ module Error = struct
           sink
           (String.concat labels ~sep:", ")
     | InvalidMultiSink sink -> Format.fprintf formatter "`%s` is not a multi sink" sink
-    | RuleCodeDuplicate code ->
-        Format.fprintf formatter "Multiple rules share the same code `%d`" code
+    | RuleCodeDuplicate { code; previous_location = None } ->
+        Format.fprintf
+          formatter
+          "Multiple rules share the same code `%d`, previous rule was defined inside the pyre \
+           binary"
```
arthaud, line 801: That would be confusing if we showed that to the user. They wouldn't know what the "pyre binary" is.
I guess this can only happen in tests right?
```suggestion
          "Multiple rules share the same code `%d`"
```

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -855,6 +882,9 @@ module Error = struct
         "code", `Int (code kind);
         "location", location;
       ]
+
+
+  let hash = Hashtbl.hash
```
arthaud, line 887: We should use `@@deriving hash` on `type t` instead. I'm not sure what this does?

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1003,16 +1028,17 @@ let from_json_list source_json_list =
...
 
     let parse ~path json =
       json_string_member ~path "name" json
       >>= fun name ->
       json_string_member ~path "message_format" json
       >>= fun message_format ->
-      json_integer_member ~path "code" json
-      >>= fun code ->
-      validate_code_uniqueness ~path code >>= fun () -> Result.Ok { name; message_format; code }
+      json_integer_member_with_location ~path "code" json
```
arthaud, line 1039: That means we use the location of the `code`. Shouldn't we use the location of the whole json object for the rule? I think this would make more sense if we want to use it latter in VSCode.

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1517,6 +1545,49 @@ let validate ({ Heap.sources; sinks; transforms; features; _ } as configuration)
     |> Result.combine_errors_unit
     |> Result.map_error ~f:List.concat
   in
+  let compare_locations previous_location location =
+    match previous_location, location with
+    | Some previous_location, Some location ->
+        JsonAst.LocationWithPath.equal previous_location location
+    | None, None -> true
+    | _ -> false
+  in
```
arthaud, line 1554: Use `Option.equal JsonAst.LocationWithPath.equal`

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1517,6 +1545,49 @@ let validate ({ Heap.sources; sinks; transforms; features; _ } as configuration)
...
+  let ensure_list_unique_with_location ~get_key ~get_error ~get_location elements =
+    let table = Hashtbl.create (module String) in
+    let seen_errors = Hash_set.create (module Error) in
+    let ensure_unique element =
+      let key = get_key element in
+      let location = get_location element in
+      match Hashtbl.find table key with
+      | None ->
+          let _ = Hashtbl.add table ~key ~data:location in
```
arthaud, line 1563: nit: since `Hashtbl.add` returns `unit` it's better to use `()` instead of `_`. 

Using `let _ = ..` is a bit dangerous because you could forget to pass a parameter and it wouldn't complain, i.e `let _ = Hashtbl.add table` compiles.

```suggestion
          let () = Hashtbl.add table ~key ~data:location in
```

source/jsonParsing.ml
```diff
@@ -76,13 +76,13 @@ module JsonAst = struct
       line: int;
       column: int;
     }
-    [@@deriving equal, show]
+    [@@deriving equal, show, compare, sexp_of]
```
arthaud, line 79: Do we really need `sexp_of`? Is it because we need it to be `Hashable`? In that case you can use `deriving hash`?

source/interprocedural_analyses/taint/taintConfiguration.ml
```diff
@@ -1517,6 +1545,49 @@ let validate ({ Heap.sources; sinks; transforms; features; _ } as configuration)
...
+            | None ->
+                { Error.path = None; location = None; kind = get_error element previous_location }
+          in
+          (* Since there might be two generated rules with the same code, when we encounter another,
+             if we don't compare, we might throw in duplicate errors.*)
+          if Hash_set.mem seen_errors error then
+            Result.Ok ()
+          else (
+            Hash_set.add seen_errors error;
+            Result.Error [error])
```
arthaud, line 1585: Could we go back to using `List.sort_and_dedup`? I would avoid hash sets in ocaml and use the "functional" approach. It's fine if we need to implement `compare` on `Error.t`.

arthaud (changes requested):

source/interprocedural_analyses/taint/taintConfiguration.ml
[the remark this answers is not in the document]
abishekvashok: Ah makes sense!

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: No, location is `None` for rules defined implicitly as I mentioned before:

> Not sure if we make it non-optional what to do for implicitly defined rules. For example, consider these defined in taintConfiguration.ml, which position and path should we use?
> 
> https://github.com/facebook/pyre-check/blob/8c639dca3a937026cf6d14130964cc485923c066/source/interprocedural_analyses/taint/taintConfiguration.ml#L515-L610

Maybe we can do something like: `Rule code `%d` cannot be used as it is reserved internally by pysa. Please use a different rule code instead`

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Ah, rules can be long, and then again the error just pertains to the `code` of a particular rule being same as another one defined before. I can change this but I strongly wanna keep it like this :)

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Thanks for teaching me that!

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: `List.dedup_and_sort` takes a tiny bit more of performance but okay!

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Thanks for that!

abishekvashok:
Thanks for the review, here's it back to you :)

source/interprocedural_analyses/taint/taintConfiguration.ml
arthaud: Those rules are in `TaintConfiguration.default`, which should only be used in tests, right?

source/interprocedural_analyses/taint/taintConfiguration.ml
arthaud: sure

arthaud (changes requested):
I would just change the error message and we should be good.

source/interprocedural_analyses/taint/taintConfiguration.ml
abishekvashok: Yes you are right. They are only used in tests. We can go with the above suggested message in that case :)

arthaud (approved):

PR #740, opened 2023-05-31
Adds verify taint config only option

abishekvashok:
Adds verify_taint_config_only option to the analyze command that just verifies taint.config files and skips the analysis. This can be useful in the future when we want to verify taint config files via the vs code extension without performing analysis or even any other sort of preprocessing.

Modifies the ocaml server and the python client to pass options for the same.

**Pre-submission checklist**
- [x] `pre-commit run`

## Test Plan
- Ran pysa with the default `taint.config` on `documentation/pysa_tutorial/exercise1/`:
Screenshot 2023-05-31 at 3 48 17 PM

- Command: `python3 -m pyre-check.client.pyre -n analyze --verify-taint-config-only`

Screenshot 2023-05-31 at 3 41 33 PM

- Modify `taint.config` to induce a TaintConfigurationError:
```json
{
  "sources": [
    {
      "name": "CustomUserControlled",
      "comment": "use to annotate user input"
    }
  ],

  "sinks": [
    {
      "name": "CodeExecution",
      "comment": "use to annotate execution of python code"
    }
  ],

  "features": [],

  "rules": [
    {
      "name": "Possible RCE:",
      "code": 5001,
      "sources": [ "CustomUserControlled" ],
      "sinks": [ "CodeExecution" ],
      "message_format": "User specified data may reach a code execution sink"
    },
    {
      "name": "test-duplicate",
      "code": 5001,
      "sources": [ "CustomUserControlled" ],
      "sinks": [ "CodeExecution" ],
      "message_format": "duplicate"
    }
  ]
}
```
Command: `python3 -m pyre-check.client.pyre -n analyze --verify-taint-config-only`

Screenshot 2023-05-31 at 3 42 49 PM

- Github actions. (pysa action was failing before this PR)

Fixes part of https://github.com/MLH-Fellowship/pyre-check/issues/82

### DIFF
--- a/source/command/analyzeCommand.ml
+++ b/source/command/analyzeCommand.ml
@@ -43,6 +43,7 @@ module AnalyzeConfiguration = struct
     maximum_tito_depth: int option;
     no_verify: bool;
     verify_dsl: bool;
+    verify_taint_config_only: bool;
     repository_root: PyrePath.t option;
     rule_filter: int list option;
     source_filter: string list option;
@@ -105,6 +106,9 @@ module AnalyzeConfiguration = struct
           let maximum_tito_depth = optional_int_member "maximum_tito_depth" json in
           let no_verify = bool_member "no_verify" ~default:false json in
           let verify_dsl = bool_member "verify_dsl" ~default:false json in
+          let verify_taint_config_only =
+            bool_member "verify_taint_config_only" ~default:false json
+          in
           let repository_root = optional_path_member "repository_root" json in
           let rule_filter = optional_list_member ~f:to_int "rule_filter" json in
           let source_filter = optional_list_member ~f:to_string "source_filter" json in
@@ -140,6 +144,7 @@ module AnalyzeConfiguration = struct
               maximum_tito_depth;
               no_verify;
               verify_dsl;
+              verify_taint_config_only;
               repository_root;
               rule_filter;
               source_filter;
@@ -204,6 +209,7 @@ module AnalyzeConfiguration = struct
         maximum_tito_depth;
         no_verify;
         verify_dsl;
+        verify_taint_config_only;
         rule_filter;
         source_filter;
         sink_filter;
@@ -260,6 +266,7 @@ module AnalyzeConfiguration = struct
       dump_call_graph;
       verify_models = not no_verify;
       verify_dsl;
+      verify_taint_config_only;
       rule_filter;
       source_filter;
       sink_filter;
@@ -305,6 +312,13 @@ let with_performance_tracking ~debug ~f =
   result
 
 
+let verify_configuration ~static_analysis_configuration () =
+  let (_ : Taint.TaintConfiguration.Heap.t) =
+    TaintAnalysis.initialize_configuration ~static_analysis_configuration
+  in
+  ()
+
+
 let run_analyze analyze_configuration =
   let { AnalyzeConfiguration.base = { CommandStartup.BaseConfiguration.source_paths; debug; _ }; _ }
     =
@@ -324,11 +338,14 @@ let run_analyze analyze_configuration =
           | _ -> true)
         ~f:(fun scheduler ->
           with_performance_tracking ~debug ~f:(fun () ->
-              TaintAnalysis.run_taint_analysis
-                ~static_analysis_configuration
-                ~build_system
-                ~scheduler
-                ());
+              if static_analysis_configuration.verify_taint_config_only then
+                verify_configuration ~static_analysis_configuration ()
+              else
+                TaintAnalysis.run_taint_analysis
+                  ~static_analysis_configuration
+                  ~build_system
+                  ~scheduler
+                  ());
           Lwt.return (ExitStatus.CheckStatus CheckCommand.ExitStatus.Ok)))
 
 
--- a/source/interprocedural_analyses/taint/taintAnalysis.ml
+++ b/source/interprocedural_analyses/taint/taintAnalysis.ml
@@ -36,10 +36,10 @@ let initialize_configuration
         _;
       }
   =
-  Log.info "Verifying model syntax and configuration.";
+  Log.info "Verifying taint configuration.";
   let timer = Timer.start () in
+  let open Core.Result in
   let taint_configuration =
-    let open Core.Result in
     TaintConfiguration.from_taint_model_paths taint_model_paths
     >>= TaintConfiguration.with_command_line_options
           ~rule_filter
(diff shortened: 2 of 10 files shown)


source/command/analyzeCommand.ml
```diff
@@ -305,6 +312,16 @@ let with_performance_tracking ~debug ~f =
...
 
 
+let verify_taint_configuration ~static_analysis_configuration () =
+  try TaintAnalysis.initialize_configuration ~static_analysis_configuration with
+  | Taint.TaintConfiguration.TaintConfigurationError _ as exn -> raise exn
+  | exn ->
+      (* The backtrace is lost if the exception is caught at the top level, because of `Lwt`. *
+         Let's print the exception here to ease debugging. *)
+      Log.log_exception "Taint analysis failed." exn (Worker.exception_backtrace exn);
+      raise exn
```
tianhan0, line 322: @arthaud Do you have a second opinion if we can do better for the logging?

source/command/analyzeCommand.ml
```diff
@@ -324,11 +341,15 @@ let run_analyze analyze_configuration =
...
           with_performance_tracking ~debug ~f:(fun () ->
-              TaintAnalysis.run_taint_analysis
-                ~static_analysis_configuration
-                ~build_system
-                ~scheduler
-                ());
+              if static_analysis_configuration.verify_taint_config_only then
+                let _ = verify_taint_configuration ~static_analysis_configuration () in
+                ()
+              else
```
tianhan0, line 347: We can do this earlier. Maybe after line 334?

Also, we can use `verify_taint_configuration ~static_analysis_configuration () |> ignore` to discard the result.

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -61,15 +61,31 @@ let initialize_configuration
           ~maximum_tito_depth
     |> TaintConfiguration.exception_on_error
   in
+  let () =
+    Statistics.performance
+      ~name:"Verified taint configuration"
+      ~phase_name:"Verified taint configuration"
+      ~timer
+      ()
+  in
```
tianhan0, line 70: Not sure if it is useful to log this, but I guess it also doesn't hurt. cc @arthaud

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -61,15 +61,31 @@ let initialize_configuration
...
+let initialize_configuration_with_models ~static_analysis_configuration =
+  Log.info "Verifying model syntax.";
+  let taint_configuration = initialize_configuration ~static_analysis_configuration in
+  let timer = Timer.start () in
   (* In order to save time, sanity check models before starting the analysis. *)
   let () =
-    ModelParser.get_model_sources ~paths:taint_model_paths
+    ModelParser.get_model_sources
+      ~paths:
+        static_analysis_configuration.Configuration.StaticAnalysis.configuration.taint_model_paths
```
tianhan0, line 82: Is there a reason to **not** use `taint_model_paths` here?

tianhan0 (changes requested):
Thanks for proposing this! I just noted a few minor things

source/command/analyzeCommand.ml
abishekvashok: Much better! Thanks for this suggestion :)

source/interprocedural_analyses/taint/taintAnalysis.ml
abishekvashok: Yeah, not much most of the time it just takes a split of a micro second.

source/interprocedural_analyses/taint/taintAnalysis.ml
abishekvashok: I redefined the function definition  to use static_analysis_configuration as we pass that as a parameter now.  We can use the shorthand `static_analysis_configuration.configuration.taint_models_path` :)

abishekvashok:
@tianhan0 thanks for the review :) Fixed all your suggestions, I hope.

source/command/analyzeCommand.ml
```diff
@@ -305,6 +312,16 @@ let with_performance_tracking ~debug ~f =
   result
 
 
+let verify_taint_configuration ~static_analysis_configuration () =
+  try TaintAnalysis.initialize_configuration ~static_analysis_configuration with
+  | Taint.TaintConfiguration.TaintConfigurationError _ as exn -> raise exn
+  | exn ->
+      (* The backtrace is lost if the exception is caught at the top level, because of `Lwt`. *
+         Let's print the exception here to ease debugging. *)
+      Log.log_exception "Taint analysis failed." exn (Worker.exception_backtrace exn);
```
tianhan0, line 321: nitpick: Since we didn't do taint analysis here, maybe change the log message to something like "The taint configuration does not pass validation"?

tianhan0 (approved):
Looks good! Just left a nitpick that hope you can address

abishekvashok:
@tianhan0

> Also, do you know why this test failed: https://github.com/facebook/pyre-check/actions/runs/5141096156/jobs/9253212045?pr=740? Is this expected?

That test has been failing for a long time and is unrelated to this PR. I think, from some tests that I did that the setup ocaml action actually uses caches which need to be invalidated now. I tried to run the action by manually setting up opam with a package manager instead of the actions and the build doesn't error out as it does in the main repo. But then without the action, setting up the switch takes a large chunk of time. I think there is a way to specify a cache key that is invalidated for a specific amount of time. I can look into this further if you want.

> Just left a nitpick that hope you can address.

Ofcourse, I can address it. Thanks for the valuable suggestions and reviews. Made the change!

source/command/analyzeCommand.ml
```diff
@@ -315,21 +335,26 @@ let run_analyze analyze_configuration =
...
-          with_performance_tracking ~debug ~f:(fun () ->
-              TaintAnalysis.run_taint_analysis
-                ~static_analysis_configuration
-                ~build_system
-                ~scheduler
-                ());
-          Lwt.return (ExitStatus.CheckStatus CheckCommand.ExitStatus.Ok)))
+  if static_analysis_configuration.verify_taint_config_only then (
+    with_performance_tracking ~debug ~f:(fun () ->
+        verify_taint_configuration ~static_analysis_configuration () |> ignore);
```
arthaud, line 340: please avoid `ignore` since it can be error prone.

For instance, it can lead to the function not actually being called when an argument is missing:
```
let foo x y = x + y

ignore (foo 1) # compiles with no warning
```

I would use `let (_ : type) = f () in ()` instead

source/command/analyzeCommand.ml
```diff
@@ -305,6 +312,19 @@ let with_performance_tracking ~debug ~f =
...
+  try TaintAnalysis.initialize_configuration ~static_analysis_configuration with
+  | Taint.TaintConfiguration.TaintConfigurationError _ as exn -> raise exn
+  | exn ->
+      (* The backtrace is lost if the exception is caught at the top level, because of `Lwt`. *
+         Let's print the exception here to ease debugging. *)
+      Log.log_exception
+        "Failed to validate taint configuration"
+        exn
+        (Worker.exception_backtrace exn);
+      raise exn
```
arthaud, line 325: Is this copy/pasted from somewhere? Can we avoid code duplication, maybe with a function?

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -36,40 +36,44 @@ let initialize_configuration
...
+        ~maximum_return_access_path_depth_after_widening
+        ~maximum_tito_collapse_depth
+        ~maximum_tito_positions
+        ~maximum_overrides_to_analyze
+        ~maximum_trace_length
+        ~maximum_tito_depth
+  |> TaintConfiguration.exception_on_error
+
+
+let initialize_configuration_with_models ~static_analysis_configuration =
```
arthaud, line 63: I would avoid this pattern of `function foo; function foo_with_bar`. Just make two separate functions, one for taint config and one for model syntax check. Use `Statistics.performance` in both with different (phase) names.

arthaud (changes requested):

source/command/analyzeCommand.ml
abishekvashok: Ah thanks for that! I will keep this in mind :)

source/interprocedural_analyses/taint/taintAnalysis.ml
abishekvashok: okay!

source/command/analyzeCommand.ml
abishekvashok: It was copy pasted - yes, moving it with-in the scheduler fixes this issue.

arthaud (approved):

PR #754, opened 2023-07-15
Fix the condition in /client/pyre.py, function run_default_command

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

After 6de35ff7004b9e69cbf66a58156a654782791ced, we use
`command_arguments.VersionKind.NONE` instead of real `None` in Python when we need to do something about version.
But, at file /client/pyre.py, function `run_default_command`, line 1550, we still use `if command_argument.version:`.
Obviously, this condition will always evaluate as `True`, because its default value is `command_arguments.VersionKind.NONE`,
equals `"none"`
So, we just change it to:
```python
if command_argument.version != command_arguments.VersionKind.NONE:
    ...
```

### DIFF
--- a/client/pyre.py
+++ b/client/pyre.py
@@ -1554,7 +1554,7 @@ def run_default_command(
     **kwargs: object,
 ) -> commands.ExitCode:
     command_argument: command_arguments.CommandArguments = context.obj["arguments"]
-    if command_argument.version:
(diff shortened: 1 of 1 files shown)


arthaud (approved):

WangGithubUser:
Great! Only #755 to resolve now and our check will return to normal.

Issue #758, opened 2023-07-21
Bootstrap script doesn't work with `fish` shell


vthemelis:
`./scripts/setup.py` doesn't work with `fish`. This is because of the different kind of output that `opam env` gives in the case of fish:

https://github.com/ocaml/opam/blob/e0a390db72dfc400edb156c306a1e2e34f4556ed/src/state/opamEnv.ml#L617-L639

PR #759, opened 2023-07-21
Fix bootstrap script for non (ba)sh script

vthemelis:
Fixes https://github.com/facebook/pyre-check/issues/758

**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Force `opam env` to give a `bash` like output.

This allows us to parse the environment variables correctly further down.

`opam env` also has the option to export the environment as s-exp but I decided to not use it this time as I would realistically add another dependency to the script.

## Test Plan

This allows me to run the bootstrap script from my fish session.

### DIFF
--- a/scripts/setup.py
+++ b/scripts/setup.py
@@ -173,6 +173,7 @@ class Setup(NamedTuple):
                 self.opam_root.as_posix(),
                 "--set-root",
                 "--set-switch",
+                "--shell=bash",
             ]
         )
         opam_environment_variables: Dict[str, str] = {}


vthemelis:
> @vthemelis I try to improve this at #761 .Could you close this if you think #761 is better?

Hi @WangGithubUser,

Thanks for looking into this.

Why do you think that we need the additional complexity? We don't capture the output of the other commands. Also, if the other commands have side effects these could now be wrong given that they could be executing `bash`-llike commands on, say, `fish`.

WangGithubUser:
> Why do you think that we need the additional complexity? We don't capture the output of the other commands.
We may use commands `env` `init` `config`  in the future.Why don't fix them now?And, this diff not only resolved the issue, but also making it easier to fix `Opam`'s command errors.We need to change all the problematic command one by one before, but now we can only change the function `get_opam_command`.

vthemelis:
But why would you need to change them? 

There's a very specific reason we need this argument for the env command and this is because we know how to parse the bash-style output from Python. Ideally we would just get the output of the command as JSON or another serialisation format that's easily parsed from Python. I don't see why we need the to force the shell on any other case.

WangGithubUser:
@vthemeli I think, make all the commands outputs same in all cases is better for future.`opam` may update in future. Make all outputs to be `bash-like` can make out developer focus more on features rather than considering whether the output will be different. And you can see, #761 's check were all passed, there isn't any side effect. Or there are some hidden efffects? If there are, please tell me.

stroxler:
I think for now this minimal change makes sense.

@WangGithubUser your change seems reasonable, but it still has some limitations because if new opam commands pop up the problem could reappear. So I think I'd like to try just setting the $SHELL env var instead as a longer-term fix; the opam docs say they will look at the SHELL var when no flag is passed

PR #767, opened 2023-07-29
Use if-elif-else instead match-case to support Python less than 3.10

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

In [commit `Fix pyre_language_server completion response parsing`, file client/language_server/code_navigation_request.py, L116-L125](https://github.com/facebook/pyre-check/commit/5db6e5d1f96645db17d4ad59a6708f56d6516c23#diff-2cbcdc81e14a25a53c1694f5cf24f4f0e4e2cc5bfb61b574f18585be88e4ece5R116-R125), we used match-case statement.
But, this statement was added in `Python 3.10`, and we are still supporting Python >= 3.8
So, we need to use if-elif-else statement instead.

## Test Plan

[The script](client/language_server/code_navigation_request.py) now can run in any version of Python that >= 3.8

* Footnote: Pyre Github Actions were failing before and is unrelated to this PR.(Actually this PR fixed the workflow `tests`)

### DIFF
--- a/client/language_server/code_navigation_request.py
+++ b/client/language_server/code_navigation_request.py
@@ -114,15 +114,14 @@ class PyreCompletionItemKind(str, enum.Enum):
     PROPERTY = "PROPERTY"
 
     def to_lsp_completion_item_kind(self) -> lsp.CompletionItemKind:
-        match self:
-            case self.SIMPLE:
-                return lsp.CompletionItemKind.TEXT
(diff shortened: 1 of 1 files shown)


WangGithubUser:
This has resolved in 446d0f68487da387b06082c4b04008ed337c2a6a, I will close this.

PR #772, opened 2023-08-03
Fix Pyre path find when init Pysa

### LINKED ISSUE #710: pyre pysa-init may be checking wrong path for filters
# Pysa Bug
**Pre-submission checklist**
[✅] I've checked the list of [common issues](https://pyre-check.org/docs/pysa-quickstart/#common-issues) and mine does not appear

**Bug description**
`pyre init-pysa` fails to find filters.

**Reproduction steps**
`pyre init-pysa`

**Expected behavior**
Shouls properly locate what I assume is the "pysa_filters" directory.

**Logs**

```
...
ƛ Couldn't infer filter directory, skipping filter import to sapp.
...
```

**Additional context**
Apologies if I'm missing something, since it seems the code has been configured this way since 2021, but init-pysa does the following:
https://github.com/facebook/pyre-check/blob/154d572dd924b3007bec8bf7478f7f5b9168c401/client/commands/initialize_pysa.py#L73-L75
This obviously doesn't work, since it is trying to look for `..."/pyre_check/taint/pyre_check"`.

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

This PR fixed #710.

## Test Plan

Pysa should init with no error message `Couldn't infer filter directory, skipping filter import to sapp.`

### DIFF
--- a/client/commands/initialize_pysa.py
+++ b/client/commands/initialize_pysa.py
@@ -73,10 +73,10 @@ def _setup_environment() -> None:
     run_infer = log.get_yes_no_input("Would you like to generate type annotations?")
     if run_infer:
         subprocess.run(["pyre", "infer", "-i"], check=True)
-    pyre_check_path = find_pysa_filters_directory()
-    if pyre_check_path is not None:
+    pysa_filter_path = find_pysa_filters_directory()
+    if pysa_filter_path is not None:
         LOG.info("Importing filters to sapp")
-        subprocess.run(["sapp", "filter", "import", pyre_check_path], check=True)
+        subprocess.run(["sapp", "filter", "import", pysa_filter_path], check=True)
     else:
         LOG.warning("Couldn't infer filter directory, skipping filter import to sapp.")
 
--- a/client/commands/initialize_pysa.py
+++ b/client/commands/initialize_pysa.py
@@ -18,7 +18,7 @@ import sys
 from pathlib import Path
 
 from .. import log
-from ..find_directories import find_taint_models_directory
+from ..find_directories import find_pyre_path
 
 from . import commands
 from .initialize import (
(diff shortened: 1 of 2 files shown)


client/commands/initialize_pysa.py
```diff
@@ -73,9 +73,8 @@ def _setup_environment() -> None:
     run_infer = log.get_yes_no_input("Would you like to generate type annotations?")
     if run_infer:
         subprocess.run(["pyre", "infer", "-i"], check=True)
-    pyre_check_path = find_taint_models_directory()
-    if pyre_check_path and os.path.isdir(pyre_check_path / "pyre_check"):
-        pyre_check_path = pyre_check_path / "pyre_check"
+    pyre_check_path = find_pyre_directory()
+    if pyre_check_path is not None:
```
arthaud, line 77: I think we should call `sapp filter` with `lib/pyre_check/pysa_filters` instead of `lib/pyre_check`.
Could you update the code to do that? We could have a `find_pysa_filters_directory()`

client/find_directories.py
```diff
@@ -301,9 +301,20 @@ def find_typeshed_search_paths(
...
+
+def find_taint_models_directory() -> Optional[Path]:
+    pyre_check_path = find_pyre_directory()
+    if pyre_check_path is not None:
+        bundled_taint_models = pyre_check_path / "taint/"
+        if bundled_taint_models.is_dir():
+            return bundled_taint_models
+        return None
+    else:
+        return None
```
arthaud, line 320: nitpick: using the "early return" pattern would make this easier to read
```suggestion
    if pyre_check_path is None:
      return None
    bundled_taint_models = pyre_check_path / "taint/"
    if bundled_taint_models.is_dir():
      return bundled_taint_models
    return None
```

arthaud (changes requested):
Thanks for the PR! See my requested changes.

arthaud (approved):

arthaud:
Thanks for the contribution!

PR #774, opened 2023-08-10
Add pre-commit hook for Pysa

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Add pre-commit hook for Pysa and improve naming of the hooks.

## Test Plan

Now users can run Pysa in pre-commit.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,5 +1,5 @@
 - id: pyre
-  name: pyre-check
+  name: Run type check for your project with Pyre
   description: "Pyre: Performant type-checking for python."
   entry: pyre check
   language: python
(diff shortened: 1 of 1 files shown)


arthaud:
Hi @WangGithubUser,

I'm personally not a fan of git hooks:
* It arbitrarily triggers on git commands (from what I know): a user might not expect pysa to run when performing git commands
* We don't use git internally so we will never run these, they might break without us (maintainers) knowing
* I would instead recommend using Github Actions: all commits trigger these runs and we get alerts on errors. Note that we already have a Github Action that runs Pysa: https://github.com/facebook/pyre-check/blob/main/.github/workflows/pysa.yml

PR #780, opened 2023-08-20
Raise the object instead of the exception type

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

We should provide `raise` the object instance, but not the type. 

## Test Plan

N/A


WangGithubUser:
Seems this PR isn't necessary, I will close it now...Python will auto change the exception type to the instance object.

Issue #773, opened 2023-08-09
Pyre couldn't find a required site-package


WangGithubUser:
# Pyre Bug

**Bug description**
Pyre can't find a Python Extension Module (.pyd) in site-package

**Reproduction steps**
Run pyre with the following `.pyre_configuration`:
```json
{
  "site_package_search_strategy": "pep561",
  "source_directories": [
    "."
  ],
  "search_path": [
    {
      "site-package": "ujson"
    }
  ],
  "taint_models_path": "Secret",
  "typeshed": "Secret"
}
```

**Expected behavior**
It should pass but not warning ``ƛ Site package does not exist: `ujson` ``

**Logs**
```
pyre check
ƛ Site package does not exist: `ujson`
ƛ No type errors found
```
**Additional context**
This may due to `pyre` do not treat `.pyd` files as a package.

WangGithubUser:
Fix suggestion: We can use `dist-info` to judge if a package exist.If a package exist, there should be a `package_name-version.dist-info` in `site-packages` directory.
![image](https://github.com/facebook/pyre-check/assets/120731947/cf0f9382-a859-4e6c-8ae0-f4f574d59286)
We can use `re` to judge if a package exist, but only add `.py` file and a folder named with the package name, just as the pseudo code below.
`client/configuration/search_path.py`:
![image](https://github.com/facebook/pyre-check/assets/120731947/128b9f23-af50-47a5-a854-f7aa1158580b)

PR #781, opened 2023-08-21
Improve site package pathing

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

A top-level package could implement by many ways, like `.so`, `.pyd` file, etc.
But, they must list in the `RECORD` file of their `dist-info`.
So, we can filter them out.

Also see https://github.com/facebook/pyre-check/issues/773#issuecomment-1672481776.

## Test Plan

Now all the implementation of the site-packages could be found.


client/configuration/search_path.py
```diff
@@ -72,9 +78,43 @@
...
+        for directory in dist_info_in_root[self.site_root]:
+            if this_pkg_filter.fullmatch(directory):
+                dist_info_path = f"{self.site_root}/{directory}"
+                break
+        else:
+            return None
+
+        not_toplevel_patterns: Tuple[re.Pattern] = (this_pkg_filter, _PYCACHE)
+
+        with open(file=f"{dist_info_path}/RECORD", mode="r") as record:
```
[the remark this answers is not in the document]
WangGithubUser: If the `dist_info_path` not defined, the func will return `None` in L100

kinto0:
Hi! It seems like your issue is site packages are not being detected? Is this on windows? Would you be able to try on WSL since we currently do not support windows?

WangGithubUser:
> Would you be able to try on WSL since we currently do not support windows?

image
(on WSL)

image
https://github.com/facebook/pyre-check/blob/baad875701eed38747cc30cad95eafb7754f991c/client/configuration/search_path.py#L76-L77
Pyre only treat a .py file as a site-package, and this PR fixed it.

client/configuration/search_path.py
```diff
@@ -19,13 +19,22 @@
...
 
 LOG: logging.Logger = logging.getLogger(__name__)
 
+dist_info_in_root: Dict[str, List[str]] = {}
+
+# pyre-fixme[5]: Globally accessible variable `_site_filter` has type `re.Pattern[str]` but no type is specified.
+_site_filter = re.compile(r".*-([0-99]\.)*dist-info")
+
+# pyre-fixme[5]: Globally accessible variable `_PYCACHE` has type `re.Pattern[str]` but no type is specified.
+_PYCACHE = re.compile("__pycache__(/)*.*")
```
WangGithubUser, line 36: No annotation for these two vars to keep compatible with Python3.8

client/backend_arguments.py
```diff
@@ -69,7 +69,8 @@ def serialize(self) -> Dict[str, object]:
         }
 
     def get_checked_directory_allowlist(self) -> Set[str]:
-        return {element.path() for element in self.elements}
+        # pyre-fixme: Incompatible return type [7]: Expected Set[str] but got Set[Optional[str]].
+        return {element.path() for element in self.elements if element is not None}
```
kinto0, line 73: element's path [can be empty](https://fburl.com/cz3a6nu7). you should also check that it's not `None`

client/backend_arguments.py
```diff
@@ -98,7 +99,8 @@ def serialize(self) -> Dict[str, object]:
...
```
kinto0, line 103: same as above

client/configuration/search_path.py
kinto0: could we at least type it as re.Pattern (without the generic)? otherwise, could we put a comment here linking to the docs and explaining why the type is omitted? https://docs.python.org/3/library/re.html#re.Pattern

client/configuration/search_path.py
```diff
@@ -72,15 +81,55 @@ class SitePackageElement(Element):
...
+        else:
+            return None
+
+        not_toplevel_patterns: Tuple[re.Pattern[str], re.Pattern[str]] = (
+            this_pkg_filter,
+            _PYCACHE,
+        )
+
+        # pyre-fixme[61]: Local variable `dist_info_path` is undefined, or not always defined.
+        with open(file=f"{dist_info_path}/RECORD", mode="r") as record:
```
kinto0, line 111: could you make sure this is always defined?

client/configuration/search_path.py
```diff
@@ -35,11 +44,11 @@ def _expand_relative_root(path: str, relative_root: str) -> str:
 
 class Element(abc.ABC):
     @abc.abstractmethod
-    def path(self) -> str:
+    def path(self) -> Union[str, None]:
```
kinto0, line 47: nit: Optional[str]

could you explain the reasoning and semantic differences for making this optional in the summary?

client/configuration/search_path.py
```diff
@@ -72,15 +81,55 @@ class SitePackageElement(Element):
...
+        # pyre-fixme[61]: Local variable `dist_info_path` is undefined, or not always defined.
+        with open(file=f"{dist_info_path}/RECORD", mode="r") as record:
+            for line in record.readlines():
+                file_name = line.split(",")[0]
+                for pattern in not_toplevel_patterns:
+                    if pattern.fullmatch(file_name):
+                        break
+                else:
+                    return file_name
+
```
kinto0, line 119: this logic seems like it could be abstracted into a separate spot. specifically, a package filter seems like a construct that should have defined semantics and might belong as a class.

client/configuration/search_path.py
```diff
@@ -72,15 +81,55 @@ class SitePackageElement(Element):
...
+        return None
+
+    def path(self) -> Union[str, None]:
+        excepted_package_path: Union[str, None] = self.package_path()
```
kinto0, line 123: what does excepted mean? did you mean expected? (few other places this is used too)

WangGithubUser: Yes, it's a typo.

client/configuration/search_path.py
WangGithubUser: > could we at least type it as re.Pattern (without the generic)? otherwise, could we put a comment here linking to the docs and explaining why the type is omitted? https://docs.python.org/3/library/re.html#re.Pattern

`re.Pattern[str]` failed in Python3.8.
https://github.com/facebook/pyre-check/actions/runs/5922502819/job/16056564658#step:5:107

Never mind, I specifed type by comment.

client/configuration/search_path.py
WangGithubUser: It is a known issue that pyre doesn't understand for-else. #791

client/configuration/search_path.py
WangGithubUser: Seems right.Will make the change :)

client/configuration/search_path.py
kinto0: hmm yeah if that's the case it would be better if you could factor out the for else loop to please the static analysis

client/configuration/search_path.py
kinto0: you also might be able to try different annotations in each branch of a `if sys.version_info <= ...` check

WangGithubUser: But specify type by comment is work in all the version of Python, why do I need to write special case for 3.8?

client/configuration/search_path.py
WangGithubUser: I don't think so.I prefer to leave this fixme there until #791 fixed since it is a bug of Pyre.

WangGithubUser:
Here is a tasklist for this PR:
1. [ ] Add a cache system for files in `site_root` or the check command will be very slow.
2. [ ] Fix error `Malformed server specification JSON. Expected string, got bool`.
3. [ ] Factor out the logic of `path` func of `SitePackageElement`.

client/configuration/search_path.py
WangGithubUser: Thanks for reviewing!Resolving conversation in favor of the tasklist below.

WangGithubUser:
@kinto0 For the second task, I guess this may because a `.so` can't detect it's annotations.But I'm not sure, could you give me more info about this?

PR #799, opened 2023-09-29
Make pyre able to read configuration from `pyproject.toml`

### LINKED ISSUE #695: Allow pyre to read configuration from pyproject.toml
# Pyre Feature Request

Having custom config files for each tool is fine. Since most tools use pyproject.toml file for configuration, if pyre supports configuration using it, it would be great.

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

This PR resolves #695.
I'm not sure we should look for `pyproject.toml` silently when `.pyre_configuration` is not valid, or only look for it when a flag provided.But, I prefer the first resoulution because the second need to provide a flag in all runs.However, I want to hear you opinion.

## Test Plan

Look for if the unittest passed.
As the unittest is not avalible this time(see footnote), I ran test in my local machine with python3.10:
image

Footnote: 
* GitHub Action workflow `Test` fails due to `match` statement is incompatible with Python3.8 and 3.9, and is not related to this PR.To make this easier to review and avoid confliting, I will rebase this PR on the top of main if the `match` statement refactored with `if-elif-else` statement but not do it in this PR.
* As a post-process, we need to delete unneeded `.pyre_configuration` at the root of this project after the nightly version of Pyre with this diff deployed.I intergrated configurations into `pyproject.toml`.

### DIFF
--- a/.pyre_configuration
+++ b/.pyre_configuration
@@ -45,6 +45,12 @@
       "site-package": "tabulate"
     },
     {
+      "site-package": "tomli"
+    },
+    {
+      "site-package": "tomli_w"
+    },
+    {
       "is_toplevel_module": true,
       "site-package": "typing_extensions"
     },
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,5 @@ tabulate
 testslide>=2.7.0
 typing_extensions
 typing_inspect
+tomli
+tomli-w
--- a/client/commands/initialize.py
+++ b/client/commands/initialize.py
@@ -266,12 +266,13 @@ def get_configuration_and_path(
 def write_configuration(
     configuration: Dict[str, Any], configuration_path: Path
 ) -> None:
-    with open(configuration_path, "wb+") as configuration_file:
-        if configuration_path.suffix.lower() == ".toml":
+    if configuration_path.suffix.lower() == ".toml":
+        with open(configuration_path, "wb+") as configuration_file:
             configuration = {"tool": {"pyre": configuration}}  # [tool.pyre] section
             tomli_w.dump(configuration, configuration_file)
-        else:
-            # assume to JSON
+    else:
+        # assume to JSON
+        with open(configuration_path, "w+") as configuration_file:
             json.dump(configuration, configuration_file, sort_keys=True, indent=2)
         configuration_file.write("\n")
 
--- a/client/commands/initialize.py
+++ b/client/commands/initialize.py
@@ -266,7 +266,7 @@ def get_configuration_and_path(
 def write_configuration(
     configuration: Dict[str, Any], configuration_path: Path
 ) -> None:
-    with open(configuration_path, "w+") as configuration_file:
+    with open(configuration_path, "wb+") as configuration_file:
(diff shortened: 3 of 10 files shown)


cclauss:
Is this still a `Draft` PR?

hparente1313:
@WangGithubUser @cclauss Is this still active? Would love to see this feature.

.pyre_configuration
```diff
@@ -48,6 +48,9 @@
     {
       "is_toplevel_module": true,
       "site-package": "typing_inspect"
+    },
+    {
+      "site-package": "tomli"
```
cclauss, line 53: `tomli` should be used only on Python < 3.11 because `tomllib` is in the Python standard library in Python >= 3.11.
* https://docs.python.org/3.11/whatsnew/3.11.html

SamDc73:
Any updates on this ?

MrTrick:
It doesn't look as though @WangGithubUser has been active on github in the last 12 months.
@cclauss perhaps this PR should be considered abandoned?

There are still people asking for this feature, (myself included) what is a good way forward?

yangdanny97:
Pyre is being rewritten right now, and we plan to support `pyproject.toml` for that but it won't be available/ready-to-use for at least several months.

In the meantime, feel free to continue with this PR, or if it's inactive someone else can make a new PR.

.pyre_configuration
WangGithubUser: I use `toml`(which is a [`site-package` listed at `.pyre_configuration`](https://github.com/facebook/pyre-check/blob/main/.pyre_configuration#L42) ) instead now, is that ok?

WangGithubUser:
@cclauss It's ready for review now(I hope).

cclauss:
I do not think the migration from [`tomli`](https://pypi.org/project/tomli) to [`toml`](https://pypi.org/project/toml) was the right choice.  The former is well-maintained and up-to-date while the latter has not been updated since 2020.

Please see https://github.com/hukkin/tomli?tab=readme-ov-file#building-a-tomlitomllib-compatibility-layer for enabling the use of the Python Standard Library module `tomllib` on Python 3.11+.

WangGithubUser:
@cclauss Done.

cclauss (commented):
Looks good to me.  @MrTrick @yangdanny97 Your reviews, please?

MrTrick (commented):
As a non-expert end user, the changes seem good to me.

cclauss:
@yangdanny97 Can we please git this merged?

requirements.txt
```diff
@@ -7,3 +7,5 @@ tabulate
 testslide>=2.7.0
 typing_extensions
 typing_inspect
+tomli
+tomli-w
```
cclauss, line 11: This file should be kept sorted to make it easy to spot missing dependencies and nearly impossible to insert duplicates.

yangdanny97:
Thanks for your contribution @WangGithubUser and thanks for the reviews @cclauss and @MrTrick

We merged all the changes minus the change to `pyre init`.

PR #801, opened 2023-10-05
Fix type error in test

WangGithubUser:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

The GitHub Action [`Tests` failed](https://github.com/facebook/pyre-check/actions/runs/6399595497/job/17371884398#step:5:630)
because [keyword ignore_cleanup_errors for `tempfile.TemporaryDirectory` is added in Python 3.10](https://github.com/python/cpython/blob/313aa861ce23e83ca64284d97c1dac234c9def7c/Misc/NEWS.d/3.10.0a7.rst?plain=1#L491-L493) and is incompatible to 3.8, 3.9
As a temp batch, we could use an equivalent form of it.After we drop 3.8 and 3.9, we could back out this PR.

## Test Plan

See if the GitHub Action passed.

### DIFF
--- a/client/commands/tests/daemon_querier_test.py
+++ b/client/commands/tests/daemon_querier_test.py
@@ -832,85 +832,88 @@ def foo(bar: int):
 
 print(foo(10))
 """
-        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmpdir:
-            tmpfile = tempfile.NamedTemporaryFile(suffix=".py", dir=tmpdir)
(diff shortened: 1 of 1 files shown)


inseokhwang:
Sorry, we went ahead and removed the param usage without the try/except clause because it's a bad practice to except errors in tests that are not intended. 

https://github.com/facebook/pyre-check/commit/157e3d91b83b3ddc885ff3306c075e58a0eaf2f9

PR #850, opened 2024-05-14
Apply Github Actions security best practices - pinned hashes

Nick2bad4u:
### Pinned Dependencies

GitHub Action tags and Docker tags are mutable. This poses a security risk. GitHub's Security Hardening guide recommends pinning actions to full length commit.

- [GitHub Security Guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)
- [The Open Source Security Foundation (OpenSSF) Security Guide](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)
- OpenSSF rates this as a: Risk: **Medium (possible compromised dependencies)**

Example in this repo: publish_website.yml uses the following which could be modified:

`      uses: JamesIves/github-pages-deploy-action@releases/v4`

This PR makes this immutable by pinning it to a full length hash.

`       uses: JamesIves/github-pages-deploy-action@ec9c88baef04b842ca6f0a132fd61c762aa6c1b0 # releases/v4`

Github says directly that:

**Pin actions to a full length commit SHA**

_Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork._

This is especially important when using third-party actions.

I plan to submit another PR (assuming this one is merged) that implements the dependabot actions dependency review that creates PRs automatically to update actions to their newest versions to make this easier to manage. (For example, if a new version of `uses: JamesIves/github-pages-deploy-action` comes out.) Although it looks like you may already have this setup/partially setup.

### Secure Dockerfiles

Pin image tags to digests in Dockerfiles. With the Docker v2 API release, it became possible to use digests in place of tags when pulling images or to use them in FROM lines in Dockerfiles.

- [The Open Source Security Foundation (OpenSSF) Security Guide](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)

This is the same thing as above, only for dockerfiles.

Lastly, if needed, I can break this down into file by file PRs.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,8 +13,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
+      - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
           python-version: "3.10"
 
--- a/.github/workflows/publish_website.yml
+++ b/.github/workflows/publish_website.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
     - name: Build the Website
       run: |
         cd documentation/website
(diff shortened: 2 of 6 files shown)


stroxler:
Test failures are pre-existing (I'm working on fixing them now); this looks good to me so I'll try to get it reviewed.

The process may be a little slower than usual because the team is traveling this week for PyCon.

Nick2bad4u:
> Test failures are pre-existing (I'm working on fixing them now); this looks good to me so I'll try to get it reviewed.
> 
> The process may be a little slower than usual because the team is traveling this week for PyCon.

No worries. Once this is merged, I will move onto the next few security fixes.

stroxler (approved):
LGTM and it's approved, aiming to merge it tonight

PR #851, opened 2024-05-14
Don't use patch for pyre in upgrade tests [2/n]

stroxler:
Summary:
This is the second of two commits removing the use of
`unittest.patch` on pyre-internal functions and classes;
instead we need to always use `patch.object` to avoid getting
errors on invalid names in github CI.

Differential Revision: D57359461


stroxler:
closing - PR failed, I need to stack more changes

PR #854, opened 2024-05-16
Remove dead `pyre-ignore` directives

stroxler:
Differential Revision: D57387917


stroxler:
A file is missing from the export,I'm trying again off a fresh internal diff

Issue #860, opened 2024-05-17
Pyre returning ERROR Check command exited with non-zero return code: 1 when running with poetry


frederiksteiner:
# Pyre Bug

**Bug description**
Pyre returning ERROR Check command exited with non-zero return code: 1 since newest release when using pyre with poetry and running the following command: 
poetry run pyre --noninteractive --debug --sequential check
Was working with version 0.9.19.

**Reproduction steps**
https://github.com/frederiksteiner/pyre-bug/tree/master
Here would be a pretty small github repo, with all the steps in the read me or also here:
Step 1: Install python 3.11.9 and poetry
Step 2: Run "poetry env use 3.11"
Step 3: Run "poetry install"
Step 4: Run "poetry run pyre --noninteractive --debug --sequential check"

**Expected behavior**
Running smoothly and returning the expected 

**Logs**
Please include any relevant logs here:

```
2024-05-17 10:34:20,118 [PID 27089] INFO No binary specified, looking for `pyre.bin` in PATH
2024-05-17 10:34:20,119 [PID 27089] INFO Pyre binary is located at `/home/fs/.cache/pypoetry/virtualenvs/bug-pyre-9tZeV5hq-py3.11/bin/pyre.bin`
2024-05-17 10:34:20,119 [PID 27089] INFO Could not determine the number of Pyre workers from configuration. Auto-set the value to 7.
2024-05-17 10:34:20,120 [PID 27089] INFO No typeshed specified, looking for it...
2024-05-17 10:34:20,120 [PID 27089] INFO Found: `/home/fs/.cache/pypoetry/virtualenvs/bug-pyre-9tZeV5hq-py3.11/lib/pyre_check/typeshed`
2024-05-17 10:34:20,120 [PID 27089] INFO Writing arguments into /tmp/pyre_arguments_aw85m6yb.json...
2024-05-17 10:34:20,120 [PID 27089] DEBUG Arguments:
{
  "source_paths": {
    "kind": "simple",
    "paths": [
      "/home/fs/code/bug-pyre/src",
      "/home/fs/code/bug-pyre$tests"
    ]
  },
  "search_paths": [
    "/home/fs/.cache/pypoetry/virtualenvs/bug-pyre-9tZeV5hq-py3.11/lib/python3.11/site-packages",
    "/home/fs/.cache/pypoetry/virtualenvs/bug-pyre-9tZeV5hq-py3.11/lib/pyre_check/typeshed/stdlib",
[...]

Issue #551, opened 2021-12-08
Pyre does not work: ƛ Check command exited with non-zero return code: 1


alega19:
# Pyre Bug

**Bug description**
Does not work, returns exit_code = 1

**Reproduction steps**
Ubuntu 18.04.6 LTS, python 3.7, current dir = /home/oleg
```
mkdir testpyre
cd testpyre
python3.7 -m venv .venv
source .venv/bin/activate
echo 'x: int = "string"' > main.py
pip install pyre-check==0.9.8
pyre init
```
I accept Default: `.`
`cat .pyre_configuration`:
```
{
  "source_directories": [
    "."
  ],
  "taint_models_path": "/home/oleg/testpyre/.venv/lib"
}
```
`pyre`:
```
ƛ No watchman binary found. 
To enable pyre incremental, you can install watchman: https://facebook.github.io/watchman/docs/install
ƛ Defaulting to non-incremental check.
ƛ Using virtual environment site-packages in search path...
ƛ '/home/oleg/testpyre/.venv/local/lib/python3.7/dist-packages' does not match any paths.
ƛ '/home/oleg/testpyre/.venv/lib/python3/dist-packages' does not match any paths.
ƛ '/home/oleg/testpyre/.venv/lib/python3.7/dist-packages' does not match any paths.
ƛ '/home/oleg/.local/lib/python3.7/site-packages' does not match any paths.
ƛ Check command exited with non-zero return code: 1.
```

I tries use flag `-n`
`pyre -n`:
```
2021-12-08 20:14:48,407 [PID 29288] WARNING No watchman binary found. 
To enable pyre incremental, you can install watchman: https://facebook.github.io/watchman/docs/install
2021-12-08 20:14:48,408 [PID 29288] WARNING Defaulting to non-incremental check.
2021-12-08 20:14:48,409 [PID 29288] WARNING Using virtual environment site-packages in search path...
2021-12-08 20:14:48,410 [PID 29288] INFO No binary specified, looking for `pyre.bin` in PATH
2021-12-08 20:14:48,410 [PID 29288] INFO Could not determine the number of Pyre workers from configuration. Auto-set the value to 1.
2021-12-08 20:14:48,410 [PID 29288] WARNING '/home/oleg/testpyre/.venv/local/lib/python3.7/dist-packages' does not match any paths.
2021-12-08 20:14:48,410 [PID 29288] WARNING '/home/oleg/testpyre/.venv/lib/python3/dist-packages' does not match any paths.
[...]

PR #863, opened 2024-05-21
Propagate stderr from pyre.bin

vthemelis:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Make it easier to diagnose issues like https://github.com/facebook/pyre-check/issues/860

## Test Plan

Ran locally and got the full error message when running pyre
```
# poetry run pyre
ƛ No watchman binary found. 
To enable pyre incremental, you can install watchman: https://facebook.github.io/watchman/docs/install
ƛ Defaulting to non-incremental check.
ƛ Path does not exist for search path: SubdirectoryElement(root='/', subdirectory='tests')
ƛ /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin)
ƛ /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin: /lib64/libc.so.6: version `GLIBC_2.29' not found (required by /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin)
ƛ /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin)
ƛ /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin)
ƛ /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /root/.cache/pypoetry/virtualenvs/bug-pyre-il7asoJj-py3.11/bin/pyre.bin)
ƛ Check command exited with non-zero return code: 1.
```

### DIFF
--- a/client/commands/check.py
+++ b/client/commands/check.py
@@ -192,11 +192,16 @@ def _run_check_command(command: Sequence[str]) -> CheckResult:
             result = subprocess.run(
                 command,
                 stdout=subprocess.PIPE,
-                stderr=log_file.file,
+                stderr=subprocess.PIPE,
                 universal_newlines=True,
                 errors="replace",
             )
             return_code = result.returncode
+
+            if result.stderr:
+                LOG.error("Error output from check command:")
+                for line in result.stderr.splitlines():
+                    LOG.error(line)
 
             # Interpretation of the return code needs to be kept in sync with
             # `source/command/checkCommand.ml`.


client/commands/check.py
```diff
@@ -192,12 +192,17 @@ def _run_check_command(command: Sequence[str]) -> CheckResult:
             result = subprocess.run(
                 command,
                 stdout=subprocess.PIPE,
-                stderr=log_file.file,
+                stderr=subprocess.PIPE,
```
stroxler, line 195: I wonder if it would make more sense to leave this as a file, but dump the file whenever the return code is nonzero. 

This might be better if the binary were ever to dump a huge amount to stderr, plus it makes it possible to cat the file if you want to check for messages as the command is running, and it would be enough to fix the issue we just saw where we weren't seeing the glibc link error.

What do you think?

stroxler: @grievejia Do you remember how this works exactly?

I'm not an expert on context managers, but what we saw was that when the pyre binary crashed with a fairly low-level stack trace (a glibc version mismatch), Pyre was never displaying the stderr output.

I wonder if the problem might actually be in the `client/log/log.py`:
```
        if <big switch on log section from pyre>
            ...
        else:
            LOG.debug(line)
```
this may mean we are doing debug-level logging of stack traces and so they are invisible by default

@vthemelis  and / or I can dig into this but wondering if you remember off the top of your head

vthemelis: Ha, indeed, that seems to be the case (which is why the same error comes up as Debug in [this older issue](https://github.com/facebook/pyre-check/issues/551)).

Looks like if the log-line doesn't have the usual `date-time SEVERITY message` format, pyre will be using `LOG.Debug` to log this line or (I think) the severity of the previously logged line.

Even this though doesn't quite make sense as I can see DEBUG messages in the output provided in https://github.com/facebook/pyre-check/issues/860 yet the libc errors do not seem to have propagated.

stroxler: Looking at this more, I think the log-watcher is relying on threads and may not finish reading OCaml's stderr; my latest guess is that in some cases when OCaml exits nonero:

- control flow passes to this thread rather than the logging thread
- this thread keeps the interpreter lock all the way through to lines 224-228
- we are hitting the `__exit__` method before the log-watcher finishes processing
- as a result it drops the tail of the ocaml stderr

I wonder if just sticking a `time.sleep(...)` before line 225 would ensure that we finish processing stderr before logging the return code and triggering `__exit__` on the log watcher.

Do you have an easy repro of the issue at hand to try it?

vthemelis: Hello, which file would you like me to add the line to?

I noticed that there are a few things that can be done to get pyre to show the errors. Things that did it for me:

* Changing `delete` to `False` here: https://github.com/facebook/pyre-check/blob/441a6cddeb4905b631079096a9329702d6fbba50/client/backend_arguments.py#L445
* OR adding a `print` statement after https://github.com/facebook/pyre-check/blob/441a6cddeb4905b631079096a9329702d6fbba50/client/backend_arguments.py#L447
* OR adding any sort of other delay in the stack really.

To me, this sounds a bit more like an issue with `communicate()` doesn't join on the logging thread.

stroxler: I agree that thread joining is the underlying problem, but given that it's a context manager I think it may be that introducing a delay is the easiest fix. I'd be fine with adding a `print` or `time.sleep` statement anywhere that fixes the issue for now, with a comment noting the issue it's addressing (could include a link to this discussion).

I slightly prefer using a delay versus `delete=False` because we probably don't want the backend file to persist if we can avoid it, it's really *only* there to pipe logs to the frontend. But it's a temp file that will get cleaned up eventually anyway, so maybe my concern is misplaced here.

@samwgoldman might have more ideas on how to join the threads in a more principled way, I'm very much not a threading expert. But I'd like to get a short-term fix in soon and we can improve on it later rather than block this PR, anything that fixes the dropped logs issue is great work and a fix we want sooner than later!

samwgoldman: I'd really prefer to avoid adding a sleep call. Can we confirm where the log is being dropped? Based on my reading of the code what should be happening:

1. pyre.bin's stderr is redirected to a log file
2. the log file is being "tail"ed by `log.log_tailer` and converted into an Iterable for each line
3. the `StreamLogger` prefixes each line and writes to stderr via the `LOG` logger

I'm not super familiar with the code in (2) and (3), but honestly it looks overcomplicated. I think it's nice to have a separate thread to read from the file, but that thread should just `readline` on the file descriptor until `EOF`. When we join that thread, the join should wait for the thread to hit that EOF.

samwgoldman: Sorry, thinking more about this, I think we should move this code to use the lower-level `subprocess.Popen` code, then read from stderr and stdout in two separate threads. This is what `subprocess.communicate` does (and what `subprocess.run` does in turn).

The thread reading from stderr can then output to the log file and transform+log to stderr. This is much nicer than redirecting to the file and then trying to also tail that file.

stroxler: After a little more looking at this, I think the real issue isn't lack of a join, since we do `join`. I think the real issue is that `StreamLogger` has two modes it runs in, a best-effort tailing mode and a proper logging mode.

The tailing mode is used in some situations where we start a pyre daemon, e.g. `start._run_in_background`, and the logger is currently optimized for that case by having `self._should_stop_reading_stream` which gets set on `__exit__`, so that whenever the main thread decides we're done we stop reading logs.

But the proper logging mode is what we want for shell commands like `pyre check`, and when in that mode we really shouldn't be using `_should_stop_reading_stream` at all, we should run to completion.

This suggests either using two classes or an attribute to differentiate the cases

stroxler: Whoops I missed Sam's second comment.

Using Popen and two threads (or even just one thread, stdout is going to be used in the foreground process anyway) should work well here too, and then the existing file-based StreamLogger could be used only for deamonized logic like `start._run_in_background`.

We can't completely rid of the file-based approach because of the need to use a file in the daemon case

PR #876, opened 2024-06-10
Added a stepLog API to reduce code duplication. 

esohel30:
This pull request introduces a new StepLogger API to simplify and standardize the logging and performance measurement process in the Pysa taint analysis codebase. By encapsulating common patterns of starting timers, logging information, and recording performance metrics, this API helps to reduce code duplication and improve maintainability. 

Questions: 

Is this the right Idea for the steplog API? 

image

IF this general Idea is correct. Please take  look at the newly created steplog.ml and stepLog.mli file and provide any feedback. 

I will then implement the feedback and update the taintAnalysis.ml file and ask for feedback once again!

### DIFF
--- a/source/stepLogger.ml
+++ b/source/stepLogger.ml
@@ -1,0 +1,24 @@
+open Core
+
+module StepLogger = struct
+  type t = {
+    start_message: string;
+    end_message: string;
+    timer: Timer.t;
+  }
+
+  let start ~start_message ~end_message =
+    let timer = Timer.start () in
+    Log.info "%s" start_message;
+    { start_message; end_message; timer }
+
+  let finish ~step_logger ~section ~integers () =
+    let time_in_seconds = Timer.stop_in_sec step_logger.timer in
+    Log.log ~section "%s: %.3fs" step_logger.end_message time_in_seconds;
+    Statistics.performance
+      ~name:step_logger.end_message
+      ~phase_name:step_logger.start_message
+      ~timer:step_logger.timer
+      ~integers
+      ()
+end
--- a/source/stepLogger.mli
+++ b/source/stepLogger.mli
@@ -1,0 +1,8 @@
+module StepLogger : sig
+    type t
+  
+    val start : start_message:string -> end_message:string -> t
+  
+    val finish : step_logger:t -> section:Log.section -> integers:(string * int) list -> unit -> unit
+  end
+  


source/stepLogger.ml
```diff
@@ -0,0 +1,24 @@
...
+module StepLogger = struct
+  type t = {
+    start_message: string;
+    end_message: string;
+    timer: Timer.t;
+  }
+
+  let start ~start_message ~end_message =
+    let timer = Timer.start () in
+    Log.info "%s" start_message;
```
arthaud, line 12: We usually have extra `...` for the first message
```suggestion
    Log.info "%s..." start_message;
```

source/stepLogger.ml
```diff
@@ -0,0 +1,24 @@
...
+    { start_message; end_message; timer }
+
+  let finish ~step_logger ~section ~integers () =
+    let time_in_seconds = Timer.stop_in_sec step_logger.timer in
+    Log.log ~section "%s: %.3fs" step_logger.end_message time_in_seconds;
```
arthaud, line 17: I think we always use `Log.info`, no need for `section` :)
```suggestion
    Log.info "%s: %.3fs" step_logger.end_message time_in_seconds;
```

arthaud (changes requested):
> Is this the right Idea for the steplog API?

Yes, that's the idea :)

Since this would only be used by Pysa, could we move this in source/interprocedural_analysis/taint/stepLogger.ml?

Also, no need to wrap it into a `module StepLogger = struct`, since the file is `stepLogger.ml` it will automatically be available as a module `StepLogger.start/finish`.

esohel30:
I updated the code according to the feedback. Please review again!

source/stepLogger.ml
```diff
@@ -0,0 +1,24 @@
+open Core
```
arthaud, line 1: Could you remove `source/stepLogger.ml` and `.mli`?

arthaud (changes requested):
Looks good!
Could you now start using this in `taintAnalysis.ml`?
This should be in the same PR IMO, as a way to test it. Otherwise we would be pushing code that's never used.

esohel30:
Finished adding steplog api to places where applicable in taintAnalysis.ml to reduce code duplication

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -330,8 +309,7 @@ let initialize_models
     ~initial_callables
   =
   let open TaintConfiguration.Heap in
-  Log.info "Parsing taint models...";
-  let timer = Timer.start () in
+  let step_logger = StepLogger.start ~start_message:"Parsing taint models..." ~end_message:"Parsed taint models" in
```
arthaud, line 312: The start message shouldn't have the `...`.
Same below for many other calls.
```suggestion
  let step_logger = StepLogger.start ~start_message:"Parsing taint models" ~end_message:"Parsed taint models" in
```

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -348,19 +326,14 @@ let initialize_models
...
-    ();
+  StepLogger.finish ~step_logger ~integers:[];
 
   let models, errors =
     match queries with
     | [] -> models, errors
     | _ ->
-        Log.info "Generating models from model queries...";
-        let timer = Timer.start () in
+        let step_logger = StepLogger.start ~start_message:"Generating models from model queries..." ~end_message:"Generatedmodels from model queries..." in
```
arthaud, line 335: same here. There is also a typo in the end message. It's somewhat important because we use the end message to query for metrics and alarms internally.
```suggestion
        let step_logger = StepLogger.start ~start_message:"Generating models from model queries" ~end_message:"Generated models from model queries" in
```

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -681,10 +620,9 @@ let run_taint_analysis
   let override_graph_shared_memory_read_only =
     Interprocedural.OverrideGraph.SharedMemory.read_only override_graph_shared_memory
   in
-  Statistics.performance ~name:"Overrides computed" ~phase_name:"Computing overrides" ~timer ();
+  StepLogger.finish ~step_logger ~integers:[];
 
-  Log.info "Indexing global constants...";
-  let timer = Timer.start () in
+  let step_logger = StepLogger.start ~start_message:"indexing global constants..." ~end_message:"Indexed global constants" in
```
arthaud, line 625: You lost an uppercase here and you changed the end message. Again, the end message is important to preserve for us :/ 
```suggestion
  let step_logger = StepLogger.start ~start_message:"Indexing global constants" ~end_message:"Finished constant propagation analysis" in
```

source/interprocedural_analyses/taint/stepLogger.ml
```diff
@@ -0,0 +1,24 @@
+open Core
+
+module StepLogger = struct
```
arthaud, line 3: This should not be needed. I think this probably doesn't compile since it doesn't match `stepLogger.mli`
```suggestion
```

source/interprocedural_analyses/taint/stepLogger.ml
```diff
@@ -0,0 +1,24 @@
...
+    end_message: string;
+    timer: Timer.t;
+  }
+
+  let start ~start_message ~end_message =
+    let timer = Timer.start () in
+    Log.info "%s..." start_message;
+    { start_message; end_message; timer }
+
+  let finish ~step_logger ~integers () =
```
arthaud, line 15: not a big deal, but just to adopt a convention:
Usually, a module that defines a `type t` and functions on that type will take that parameter as the last (or sometimes first) parameter (and without using named parameters).
For instance, you wouldn't use `List.concat ~list` or `List.append ~list ~element`. That would break a lot of potential for chaining.

TL, DR: I would suggest something like:
```suggestion
  let finish ~integers step_logger =
```
This also gets rid of the `()`, which is a nice benefit.

arthaud (changes requested):

esohel30:
All feedback has been applied and code is ready for a final review. I double checked but as there is a lot of code I probably missed a thing or two again. Sorry in advanced and thank you for the great feedback! I am really learning a lot and feeling more confident.

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -66,32 +65,19 @@ let initialize_and_verify_configuration
...
     |> TaintConfiguration.exception_on_error
   in
-  let () =
-    Statistics.performance
-      ~name:"Initialized and verified taint configuration"
-      ~phase_name:"Initializing and verifying taint configuration"
-      ~timer
-      ()
-  in
+  StepLogger.finish ~step_logger ~integers:[];
```
arthaud, line 68: This doesn't compile, it should be:
```suggestion
  StepLogger.finish ~integers:[] step_logger;
```
I will fix it myself so we don't take another week to merge this.
In the future, please run `make test` before submitting :)

arthaud (approved):

source/interprocedural_analyses/taint/taintAnalysis.ml
```diff
@@ -348,19 +326,14 @@ let initialize_models
       ~definitions:(Some definitions_hashset)
       ~stubs:(Interprocedural.Target.HashsetSharedMemory.read_only stubs_shared_memory)
   in
-  Statistics.performance
-    ~name:"Parsed taint models"
-    ~phase_name:"Parsing taint models"
-    ~timer
-    ~integers:["models", Registry.size models; "queries", List.length queries]
```
arthaud, line 355: We should keep this too.

Issue #869, opened 2024-05-31
[pysa][MLH] Add a show to Registry


arthaud:
When debugging, it can be useful to print the content of a `Registry`:
https://github.com/facebook/pyre-check/blob/main/source/interprocedural/fixpointAnalysis.ml#L143

Unfortunately, it does not currently implement `pp` or `show`. Let's add those.

Note that we can use something like this to pretty print a map-like structure:
```ocaml
let pp_map pp_value formatter map =
    let pp_pairs pp_value formatter pairs =
      let pp_pair formatter key value =
        Format.fprintf formatter "@,%s -> [%a]" key pp_value value
      in
      iter (pp_pair formatter) pairs
    in
    Format.fprintf formatter "{@[<v 2>%a@]@,}" (pp_pairs pp_value) map

let show ~pp_value map = Format.asprintf "%a" (pp_map pp_value) map
```

PR #877, opened 2024-06-10
enhanced the registry module to support pretty printing and string co…

esohel30:
#869 

This pull request adds pp and show functions to the Registry module to enable pretty-printing of the map contents for easier debugging and logging.

Changes:

**Added pp Function:**

The pp function formats the map's key-value pairs for pretty-printing.
Utilizes Format.fprintf for formatting and printing the map contents in a readable format.

**Added show Function:**

The show function converts the map to a string representation using the pp function.

Uses Format.asprintf to generate the string representation of the map.

**Modified File:**

registry.ml

**The pull request is not fully done. Please provide any feedback on if I have the general understanding of the Issue. I will then implement the feedback and attempt to tackle the entire issue.**

### DIFF
--- a/source/interprocedural_analyses/taint/registry.ml
+++ b/source/interprocedural_analyses/taint/registry.ml
@@ -21,7 +21,7 @@ let targets_with_mode models ~mode =
   in
   fold ~init:[] ~f:collect models
 
-
+  
 let skip_overrides models =
   targets_with_mode models ~mode:Model.Mode.SkipOverrides
   |> List.filter ~f:Target.is_function_or_method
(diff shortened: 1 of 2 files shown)


source/interprocedural_analyses/taint/registry.ml
```diff
@@ -38,3 +38,15 @@ let skip_analysis models =
 
 
 let entrypoints models = targets_with_mode models ~mode:Model.Mode.Entrypoint
+
+let pp pp_value formatter map =
+  let pp_pairs formatter pairs =
+    let pp_pair formatter (key, value) =
+      Format.fprintf formatter "@[<hv 2>%a -> %a@]" Target.pp key pp_value value
```
arthaud, line 45: We want to use `Model.pp` here instead of `pp_value`

source/interprocedural_analyses/taint/registry.ml
```diff
@@ -38,3 +38,15 @@ let skip_analysis models =
...
+    in
+    Format.pp_print_list pp_pair formatter (Target.Map.bindings map)
+  in
+  Format.fprintf formatter "@[<v 2>{%a}@]" pp_pairs
+
+let show pp_value map =
+  Format.asprintf "%a" (pp pp_value) map
```
arthaud, line 52: We can make this a bit more concise

```suggestion
let show =
  Format.asprintf "%a" pp
```

arthaud (changes requested):
That's the idea :)

Could we define this directly in source/interprocedural/fixpoint.ml, in the Registry module? That file defines a generic "interprocedural fixpoint", and the taint analysis is just an instantiation. Since your pp/show functions are generic, they can go there directly :) 

Also please demonstrate that it works by pretty-printing a registry. You should be able do to that within a test.

PR #886, opened 2024-06-19
Draft PR for Pysa Fuzzer 

esohel30:
Draft PR for the pysa fuzzer project 🚀


arthaud:
Cool, this is a good start.

A few comments:
* Could we use type annotations in the scripts? See https://docs.python.org/3/library/typing.html. We try to annotate most of our python code at Meta. It should be fairly easy here, most arguments are `str` and the return value of most functions is also `str`
* I think this is not generating valid code, since a variable might be used before it is defined. Could you find a way to prevent that problem?
* There are many patterns that cannot be generated because `generate_complex_expression` is not truly recursive. For instance, `generate_if_statement` uses `generate_assignment` which itself uses `generate_variable_expression`. It will never generate a `if` condition that contains a function call, for instance. Do you have ideas on how to improve that? I can help if needed.
* Instead of implementing your own `indent`, just use `textwrap.indent` from the standard library
* Finally, most importantly: in the end, this always generate `sink(source())`. Pysa will likely find this 99.99% of the time since the whole code before that doesn't really matter. What we need is the source to actually flow through the randomly generated statements. Then, the hard part is to either: only generate code where the flow is valid, or somehow have a way to execute the code to verify if the flow is valid (to see if Pysa agrees with it).

esohel30:
Enhanced Pysa fuzzer by adding type annotations, ensuring variables are defined before use, making expression generation truly recursive, and using textwrap.indent for better code formatting. Added a defined_variables set to track declared variables, improving code validity. Despite these improvements, the fuzzer is still far from perfect and requires further refinement to enhance code generation diversity and flow validation. Will continue to work on it extensively! Might even approach it in a different manner now that I have been playing around with a it for a bit.

arthaud:
Cool, another round of feedback:

* I believe this never generates a valid flow, since `sink_var` is always the result of calling `get_next_variable` called **after** generating the expression, so that variable is never assigned. To fix that, you need the whole code generation to know the `source_var` and `sink_var` and have an assignment at some point.
* There is a bug where this can generate `x = x` since `to_var` is created before the right hand side expression is generated
* What if `get_next_variable` runs out of variable? We should probably use variable names such as `a`, .., `z`, `aa`, .., `az`, etc.
* Can we find a way to avoid using mutable global variables (e.g, `defined_variables` and `current_var_index`)? This is a problem if we want to generate several tests by calling `generate_source_to_sink` multiple times. A solution would be to use a class with attributes.
* It is still not fully recursive: `generate_arithmetic_expression` is never called, and `generate_assignment` only uses `generate_variable_expression`. This will never generate `x = foo()` or `x = y + z + 1`. Here is an idea: you should first differentiate between expressions (return a value) and statements. The current `generate_complex_expression` generates statements, not expressions. Then, the actual `generate_expression` function should be recursive and accept a maximum depth. Pick a number, then if it's 0: generate a constant, if it's 1, generate an addition, by calling `generate_expression` recursively, if it's 2, generate a substraction, etc. if it's some number N, generate a function call, etc.
* It would be interesting to also generate expressions with lists (creating lists, adding an element to a list, accessing a list item) and dictionaries (dict).

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+import textwrap
+
+def generate_variable_names():
+    names = []
+    for i in range(1, 3):  
+        for first in range(ord('a'), ord('z') + 1):
+            if i == 1:
+                names.append(chr(first))
+            else:
+                for second in range(ord('a'), ord('z') + 1):
```
arthaud, line 11: We can make this easier with `itertools.product` and `string.ascii_lowercase`, something like:
```
names = string.ascii_lowercase
names.extend(''.join(word) for word in itertools.product(string.ascii_lowercase, repeat=2))
```

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+                    if chr(first) + chr(second) != 'as':
```
arthaud, line 12: I understand that we have to avoid `as` because it's a reserved keyword. I think this deserves a comment. I would also remove it at the end, to avoid it being nested in all the loops
```suggestion
  names.remove('as')
```

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+                        names.append(chr(first) + chr(second))
+    return names
+
+variables = generate_variable_names()
+current_var = 0
```
arthaud, line 17: Avoid global variables, please!

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+def generate_if_else_elif() -> str:
+    index = get_next_idx()
+    curr_var = variables[index]
+    prev_var = variables[index - 1]
+
+    conditions = [
+        (f"if {prev_var} == '{random.randint(1, 10)}':", f"{curr_var} = {prev_var} + ' condition1'"),
+        (f"elif {prev_var} == '{random.randint(11, 20)}':", f"{curr_var} = {prev_var} + ' condition2'"),
+        (f"else:", f"{curr_var} = {prev_var} + ' condition3'")
+    ]
```
arthaud, line 160: We could make this more interesting, by having a single branch doing the `curr_var = prev_var + condition` that leads to the flow, and the other branch using variables that wouldn't be tainted. for instance, you could initialize a few variables to a string before the if/elif/else and use those like `curr_var = non_tainted_var + condition`.

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+        ("sin", f"{curr_var} = str(math.sin(float({prev_var})))"),
+        ("cos", f"{curr_var} = str(math.cos(float({prev_var})))"),
+        ("tan", f"{curr_var} = str(math.tan(float({prev_var})))")
+    ]
+
+    # Select a random operation
+    operation = random.choice(operations)
+    
+    # Import math if not already imported
+    import_math = "import math"
```
arthaud, line 185: might as well just import math just once at the beginning of the generated code.

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+    curr_var = variables[index]
+    prev_var = variables[index - 1]
+
+    # Choose a random mathematical operation from the math library
+    operations = [
+        ("sqrt", f"{curr_var} = str(math.sqrt(int({prev_var})))"),
+        ("log", f"{curr_var} = str(math.log(int({prev_var}) + 1))"),  # log(x) requires x > 0
+        ("sin", f"{curr_var} = str(math.sin(float({prev_var})))"),
+        ("cos", f"{curr_var} = str(math.cos(float({prev_var})))"),
+        ("tan", f"{curr_var} = str(math.tan(float({prev_var})))")
```
arthaud, line 178: there is no point having that first element
```suggestion
        f"{curr_var} = str(math.sqrt(int({prev_var})))",
        f"{curr_var} = str(math.log(int({prev_var}) + 1))",  # log(x) requires x > 0
        f"{curr_var} = str(math.sin(float({prev_var})))",
        f"{curr_var} = str(math.cos(float({prev_var})))",
        f"{curr_var} = str(math.tan(float({prev_var})))"

```

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+def generate_addition() -> str:
+    index = get_next_idx()
+    curr_var = variables[index]
+    prev_var = variables[index - 1]
+    return f"{curr_var} = {prev_var} + '{random.randint(1, 9)}'"
+
+def generate_for_loop() -> str:
+    index = get_next_idx()
+    curr_var = variables[index]
+    prev_var = variables[index - 1]
```
arthaud, line 45: That seems like a common pattern.
Why not have two functions, `generate_new_variable()` and `get_last_variable()`?

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,253 @@
...
+    current_var += 1
+    return temp
+
+def generate_source() -> str:
+    index = get_next_idx()
+    curr_var = variables[index]
+    return f"import random\n{curr_var} = input()"
+
+def generate_sink() -> str: 
+    index = get_next_idx()
```
arthaud, line 32: we don't really need to generate a new variable/index since we don't use it.

arthaud (changes requested):

scripts/pysa_fuzzer/fuzzer.py
```diff
@@ -0,0 +1,130 @@
...
+        self.current_var_index = 0
+
+    def indent(self, text: str, spaces: int = 4) -> str:
+        return textwrap.indent(text, ' ' * spaces)
+
+    def generate_arithmetic_expression(self, depth: int) -> str:
+        if depth == 0:
+            return str(random.randint(0, 10))
+        op = random.choice(['+', '-', '*', '/'])
+        return f"({self.generate_arithmetic_expression(depth - 1)} {op} {self.generate_arithmetic_expression(depth - 1)})"
```
arthaud, line 20: This is still not truly recursive. This should call `generate_expression`.
same thing below.

scripts/pysa_fuzzer/fuzzer.py
```diff
@@ -0,0 +1,130 @@
...
+            lambda: self.generate_while_loop(depth, from_var),
+            lambda: self.generate_try_except(depth, from_var),
+            lambda: self.generate_function_definition(depth, from_var)
+        ]
+        random.shuffle(parts)
+        return '\n'.join([part() for part in parts[:random.randint(2, 3)]])
+
+    def generate_source_to_sink(self) -> str:
+        self.current_var_index = 0  # Reset the index for each new generation
+        self.defined_variables.clear()
```
arthaud, line 120: That's not really necessary. We should just instantiate a new `CodeGenerator` each time, so the state is properly initialized.

scripts/pysa_fuzzer/fuzzer.py
```diff
@@ -0,0 +1,130 @@
...
+            lambda: self.generate_assignment(from_var),
+            lambda: self.generate_if_statement(depth, from_var),
+            lambda: self.generate_for_loop(depth, from_var),
+            lambda: self.generate_while_loop(depth, from_var),
+            lambda: self.generate_try_except(depth, from_var),
+            lambda: self.generate_function_definition(depth, from_var),
+        ]
+        return random.choice(choices)()
+
+    def generate_complex_expression(self, depth: int, from_var: str) -> str:
```
arthaud, line 103: should be renamed `generate_statements`?

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,219 @@
...
+        names = [name for name in names if name not in reserved_keywords]
+        return names
+
+    def generate_new_variable(self):
+        index = self.current_var
+        self.current_var += 1
+        return self.variables[index]
+
+    def get_last_variable(self):
+        return self.variables[self.current_var - 2]
```
arthaud, line 26: Well, `current_var - 2` is not really the last one, is it?
We should really just use `current_var - 1` and you should call `get_last_variable()` **before** calling `generate_new_variables()` in methods below.

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,219 @@
+import random
+import textwrap
+import itertools
+import string
+
+class CodeGenerator:
+    def __init__(self):
```
arthaud, line 7: nitpick: all methods that don't return anything should be annotated with `-> None:`
```suggestion
    def __init__(self) -> None:
```

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def get_last_variable(self):
+        return self.variables[self.current_var - 2]
+    
+    def generate_source(self) -> str:
+        curr_var = self.generate_new_variable()
+        return f"import random\nimport math\n{curr_var} = input()"
+
+    def generate_sink(self) -> str: 
+        curr_var = self.generate_new_variable()
```
arthaud, line 33: this is unused, and only necessary because `get_last_variable` is buggy (see comment above)

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,219 @@
...
+        operations = [
+            ("sqrt", f"{curr_var} = str(math.sqrt(int({prev_var})))"),
+            ("log", f"{curr_var} = str(math.log(int({prev_var}) + 1))"),  # log(x) requires x > 0
+            ("sin", f"{curr_var} = str(math.sin(float({prev_var})))"),
+            ("cos", f"{curr_var} = str(math.cos(float({prev_var})))"),
+            ("tan", f"{curr_var} = str(math.tan(float({prev_var})))")
+        ]
+
+        # Select a random operation
+        operation = random.choice(operations)
```
arthaud, line 161: This method doesn't return anything?

arthaud (changes requested):

alexkassil:
Hey @esohel30 , so the idea for this project is to automatically find false negatives - ie flows that should be security issues, but for whatever reason pysa doesn't find it.

The way to do this is to generate increasingly complex valid flows for pysa to find. Everything generated should be a valid security issue.

https://github.com/facebook/pyre-check/tree/main/source/interprocedural_analyses/taint/test/integration take a look at the tests here (in the .py files). 

Here's an example: https://github.com/facebook/pyre-check/blob/main/source/interprocedural_analyses/taint/test/integration/source_sink_flow.py 
```
from builtins import _test_sink, _test_source


def bar():
    return _test_source()


def qux(arg):
    _test_sink(arg)


def bad(ok, arg):
    qux(arg)


def some_source():
    return bar()


def match_flows():
    x = some_source()
    bad(5, x)
```

One issue in this file is the flow in `match_flows() -> bad()`

One way to always generate valid issues is to start with `_test_sink(_test_source())`, and then pick operations that make it so the flow is one more hop away.

For example, let's say you add 3 functionalities of mutations to the fuzzer:
1. Extra variable
2. Function call
3. if statement else clause

And now you randomly pick from those 3 elements 4 times to get [1, 2, 3, 2].

Applying those mutations step by step gets you:
```
f1():
  x = test_source()
  _test_sink(x)
```

```
def f2():
  x = test_source()
  f2(x)

def f1(x):
  _test_sink(x)
```

```
def f2(cond):
  x = test_source()
  if cond:
    pass
  else:
    f2(x)

def f1(x):
  _test_sink(x)
```

```
def f3(x, cond):
    x = test_source()
    f2(x, cond)

def f2(x, cond):
  if cond:
    pass
  else:
    f2(x)

def f1(x):
  _test_sink(x)
```

Continuing adding more and more single hop transformations that preserve the flow will make the fuzzer be able to generate all the valid flows present in https://github.com/facebook/pyre-check/tree/main/source/interprocedural_analyses/taint/test/integration - I think for simplicity do not worry about any modelling other than `_test_source` and `_test_sink`

scripts/pysa_fuzzer/explanation.md
```diff
@@ -0,0 +1,56 @@
...
+    c = ((g / g) * (b * b))
+for c in range(2, 5):
+    d = ((a / c) * (c - c))
+e = ((g - d) * (e - e))
+def foo(a, e, b):
+    d = ((g / d) / (b + b))
+qux((5 + 0))
+while (2 > 10):
+    c = ((e - c) * (a / a))
+log_output(get_data())
```
alexkassil, line 51: I don't understand this example, the only "flow" would be `log_output(get_data())`, all the other code isn't doing anything, no?

scripts/pysa_fuzzer/explanation.md
```diff
@@ -0,0 +1,56 @@
+# Program Explanation
+Pysa Fuzzer 
+
+## Purpose
+The program generates random Python code snippets that mimic different parts of a program, including assignments, loops, function calls, conditionals, and more. It's designed to create complex, varied code automatically.
+
+## Sources and Sinks
+- **Sources**: These are functions that simulate getting input from a user or another part of a system (e.g., `user_input()`, `get_data()`, `fetch_value()`).
+- **Sinks**: These are functions that represent places where data is used or outputted, potentially in a risky way (e.g., `print`, `send_data`, `log_output`).
```
alexkassil, line 9: just have `_test_source` and `_test_sink`

scripts/pysa_fuzzer/example_generation.py
```diff
@@ -0,0 +1,104 @@
+import random
+import math
+a = input()
```
alexkassil, line 3: What you generate should roughly match the format of our integation tests here: https://github.com/facebook/pyre-check/tree/main/source/interprocedural_analyses/taint/test/integration

Ideally it is a file with some function calls and 1 valid flow per file

scripts/pysa_fuzzer/fuzzer.py
```diff
@@ -0,0 +1,130 @@
...
+            return str(random.randint(0, 10))
+        op = random.choice(['+', '-', '*', '/'])
+        return f"({self.generate_arithmetic_expression(depth - 1)} {op} {self.generate_arithmetic_expression(depth - 1)})"
+
+    def generate_variable_expression(self, depth: int) -> str:
+        if depth == 0:
+            var = random.choice(list(self.defined_variables))
+            return var
+        op = random.choice(['+', '-', '*', '/'])
+        return f"({self.generate_variable_expression(depth - 1)} {op} {self.generate_variable_expression(depth - 1)})"
```
alexkassil, line 27: Working directly with stringified source code will get really tedious. I recommend working on the AST directly https://docs.python.org/3/library/ast.html for codegen.

alexkassil (changes requested):
Looking at the code, here is how I recommend you change it high level going forward:

1. Look through https://github.com/facebook/pyre-check/tree/main/source/interprocedural_analyses/taint/test/integration and pick a few simpler cases to try to model with the fuzzer. Say somewhere around 5 cases
2. If you exhaustively generate all 4 length mutations, that's 5*5*5*5, or 625 valid but different pysa flows. 
3. run pysa on all of them and validate pysa catches those errors

scripts/pysa_fuzzer/explanation.md
arthaud: This was generated by the initial version, now the code generation is better.

scripts/pysa_fuzzer/fuzzer.py
arthaud: While I generally agree, I think it is not a high priority to address that for now.

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,294 @@
...
+        sink_func = f"def {sink_func_name}(x):\n    print(x)"
+        indirect_call_func = f"def {indirect_call_func_name}(f, arg):\n    return f(arg)"
+        data_assignment = f"{data_var} = {source_func_name}()"
+        call = f"{indirect_call_func_name}({sink_func_name}, {data_var})"
+
+        return f"{source_func}\n\n{sink_func}\n\n{indirect_call_func}\n\n{data_assignment}\n\n{call}"
+
+
+
+    def generate_statements(self, x: int) -> str:
```
arthaud, line 253: nit: rename `x` into `number_statements`

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def generate_if_else_elif(self) -> str:
+        prev_var = self.get_last_variable()
+        curr_var = self.generate_new_variable()
+
+        conditions = [
+            (f"if {prev_var} == '{random.randint(1, 10)}':", f"{curr_var} = {prev_var} + ' c1'"),
+            (f"elif {prev_var} == '{random.randint(11, 20)}':", f"{curr_var} = {prev_var} + ' c2'"),
+            (f"else:", f"{curr_var} = {prev_var} + ' c3'")
+        ]
```
arthaud, line 199: You haven't addressed my previous comment: it would be better if only one branch used `prev_var`. because right now all branches have a valid flow, this makes it really easy for pysa.

scripts/pysa_fuzzer/hinder_static_analysis.py
```diff
@@ -0,0 +1,134 @@
...
+
+    data = source()
+    indirect_call(sink, data)
+
+# 2. Use of exec and eval
+def example_exec_eval():
+    source_code = "a = input()"
+    sink_code = "print(a)"
+    exec(source_code)
+    exec(sink_code)
```
arthaud, line 22: It is a known false-negative, so there is no point generating this kind of code. We already know about the problem.

scripts/pysa_fuzzer/hinder_static_analysis.py
```diff
@@ -0,0 +1,134 @@
...
+    def source():
+        global global_data
+        global_data = input()
+
+    def sink():
+        global global_data
+        print(global_data)
+
+    source()
+    sink()
```
arthaud, line 80: This is another known false negative, so I would say it's not really interesting using this.

scripts/pysa_fuzzer/Makefile
```diff
@@ -0,0 +1,31 @@
+# Generate Python files
```
arthaud, line 1: Can we rewrite this as a python script?
Sure, Makefiles could do the job here, but there are very limited.
They also force us to use bash, which is a very bad as a programming language.
For instance, the fact that you had to use another python script `find_undetected_files.py` shows the limit of the Makefile.

scripts/pysa_fuzzer/Makefile
```diff
@@ -0,0 +1,31 @@
...
+generate:
+	python generate_code.py
+
+# Set up Python virtual environment and install dependencies
+setup:
+	python3 -m venv tutorial
+	. tutorial/bin/activate && \
+	pip3 install pyre-check fb-sapp django-stubs
```
arthaud, line 9: Can we just assume the user installed pyre? It seems wasteful to re-install pyre everything I want to run that script.
Also, we shouldn't need fb-sapp and django-stubs here.

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,136 @@
+import subprocess
+import os
+import json
+import argparse
+from fuzzer2 import CodeGenerator
+
+def run_command(command):
+    process = subprocess.Popen(command, shell=True)
```
arthaud, line 8: We try to avoid `shell=True` since it can lead to security vulnerabilities. Obviously this script is not likely to lead to vulnerabilities, but for the sake of consistency, I would still avoid `shell=True` here.
Please also use `subprocess.run`, we should only use `subprocess.Popen` for the most complex use case.
And also, use `check=True`, otherwise you are ignoring errors.

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,136 @@
...
+        generated_code = generator.generate_statements(x)
+        generator.reset()
+        filename = os.path.join(output_dir, f'{i}.py')
+        with open(filename, 'w') as file:
+            file.write(generated_code)
+        print(f"Generated {filename}")
+
+def setup_virtual_environment():
+    run_command('python3 -m venv tutorial')
+    run_command('. tutorial/bin/activate && pip3 install pyre-check')
```
arthaud, line 31: cf previous comment, I think we should just assume the user already installed pyre before running the script. For instance, this will not work with our internal setup.

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,136 @@
...
+
+def run_command(command):
+    process = subprocess.Popen(command, shell=True)
+    process.communicate()
+
+def generate_python_files():
+    generator = CodeGenerator()
+    num_files = 100  # Change this number to generate a different amount of files
+    x = 20  # Change this number to generate a different amount of functions
+    output_dir = 'generated_files'
```
arthaud, line 15: These parameters could be command line arguments

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Create the directory if it doesn't exist
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+
+    for i in range(1, num_files + 1):
+        generated_code = generator.generate_statements(x)
+        generator.reset()
+        filename = os.path.join(output_dir, f'{i}.py')
```
arthaud, line 24: Python file names should technically be valid module names, and `1`, `2`, etc. are not valid module names.
Maybe use `test_{i}.py`?

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,136 @@
...
+        with open(filename, 'w') as file:
+            file.write(generated_code)
+        print(f"Generated {filename}")
+
+def setup_virtual_environment():
+    run_command('python3 -m venv tutorial')
+    run_command('. tutorial/bin/activate && pip3 install pyre-check')
+
+def configure_and_analyze():
+    os.makedirs('generated_files', exist_ok=True)
```
arthaud, line 34: We already do that line 18/19. Maybe move it in a function `create_test_directory` or something.

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,136 @@
...
+                    "code": 5001,
+                    "sources": [ "CustomUserControlled" ],
+                    "sinks": [ "CodeExecution" ],
+                    "message_format": "User specified data may reach a code execution sink"
+                }
+            ]
+        }
+        json.dump(taint_config, taint_config_file, indent=2)
+
+    run_command('. tutorial/bin/activate && cd generated_files && pyre analyze > analysis_output.tmp')
```
arthaud, line 70: use `pyre -n analyze` to get more verbose logs by default, which is useful for debugging.
Also: it's a json file, so use `.json` instead of `.tmp`?

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,317 @@
...
+        return self.variables[self.current_var - 1]
+
+    def reset(self) -> None:
+        self.current_var = 0
+
+    def generate_import_statements(self) -> str:
+        return "import random\nimport math\n"
+    
+    def generate_source(self) -> str:
+        import_statements = self.generate_import_statements()
```
arthaud, line 36: I would call this directly from `generate_statements`. There is no strong reason to have it here.

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,317 @@
...
+
+        source_func = f"def {source_func_name}():\n    return input()"
+        sink_func = f"def {sink_func_name}(x):\n    print(x)"
+        indirect_call_func = f"def {indirect_call_func_name}(f, arg):\n    return f(arg)"
+        data_assignment = f"{data_var} = {source_func_name}()"
+        call = f"{indirect_call_func_name}({sink_func_name}, {data_var})"
+
+        return f"{source_func}\n\n{sink_func}\n\n{indirect_call_func}\n\n{data_assignment}\n\n{call}"
+
+    def generate_linked_list_pattern(self) -> str:
```
arthaud, line 258: As discussed offline, this is a known false negative, so no point generating this.

arthaud (changes requested):

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,123 @@
...
+    with open('../analysis_output.tmp', 'r') as file:
+        analysis_output = json.load(file)
+
+    # Extract the list of files where the flow has been detected
+    detected_files = set()
+    for entry in analysis_output:
+        detected_files.add(os.path.basename(entry['path']))  # Normalize the path to get the filename only
+
+    # Generate the full list of files (from test_1.py to test_100.py)
+    all_files = {f"test_{i}.py" for i in range(1, 101)}
```
arthaud, line 81: Here we should use `num_files`

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # Find the files where the flow has not been detected
+    undetected_files = all_files - detected_files
+
+    # Output the list of undetected files
+    print("Files where the flow has not been detected:")
+    for file in sorted(undetected_files, key=lambda x: int(x.split('_')[1].split('.')[0])):
```
arthaud, line 88: We can do slightly better to avoid the `splits`:

```suggestion
    for file in sorted(undetected_files, key=lambda x: int(x[len('test_'):-len('.py')])):
```

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,123 @@
...
+    undetected_count = len(undetected_files)
+    undetected_percentage = (undetected_count / total_files) * 100
+    print(f"Flow has not been detected in {undetected_percentage:.2f}% of the files")
+
+def clean_up():
+    # will work later
+    subprocess.run(['rm', 'sources_sinks.pysa'], check=True)
+    subprocess.run(['rm', 'taint.config'], check=True)
+    subprocess.run(['rm', '.pyre_configuration'], check=True)
+    subprocess.run(['rm', 'analysis_output.tmp'], check=True)
```
arthaud, line 102: Use `os.remove` and `shutil.rmtree` instead of `subprocess(['rm', ..])`.

scripts/pysa_fuzzer/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+- **`--num_files`**: Specifies the number of files to generate (default is 100).
+- **`--x`**: Specifies the number of functions to generate in each file (default is 20).
+
+### Examples
+
+#### 1. Generate Files, Configure, and Run Analysis
+
+```bash
+python script_name.py all --num_files 50 --x 10
```
arthaud, line 33:
```suggestion
python build_script.py all --num_files 50 --x 10

```

scripts/pysa_fuzzer/build_script.py
```diff
@@ -0,0 +1,123 @@
+import subprocess
```
arthaud, line 1: naming: I would call this script `run.py` and `fuzzerX.py` would be called `code_generator.py`

scripts/pysa_fuzzer/fuzzer2.py
```diff
@@ -0,0 +1,279 @@
...
+    def generate_if_else_elif(self) -> str:
+        prev_var = self.get_last_variable()
+        non_tainted_var1 = self.generate_new_variable()
+        non_tainted_var2 = self.generate_new_variable()
+        curr_var = self.generate_new_variable()
+
+        conditions = [
+            (f"if {prev_var} == {prev_var}:", f"{curr_var} = {prev_var} + 'c1'"),
+            (f"elif {prev_var} == '{random.randint(11, 20)}':", f"{curr_var} = {non_tainted_var1} + 'c2'"),
+            (f"else:", f"{curr_var} = {non_tainted_var2} + 'c3'")
```
arthaud, line 206: good, but now the flow is always in the `if`.
Let's pick a random number between 1 and 3 to decide in which branch the valid flow should be.

arthaud (changes requested):

scripts/pysa_fuzzer/run.py
```diff
@@ -0,0 +1,148 @@
...
+            with open(output_file, 'w') as file:
+                file.write(result.stdout)
+    except subprocess.CalledProcessError as e:
+        logging.error(f"Command '{command}' failed with error: {e.stderr}")
+
+def generate_python_files(num_files, num_statements):
+    generator = CodeGenerator()
+    output_dir = Path('generated_files')
+
+    # Create the directory if it doesn't exist
```
tianhan0, line 25: nitpick: This comment doesn't add much value, since `mkdir` and `exist_ok` makes it clear what this is doing. We could remove this.

scripts/pysa_fuzzer/run.py
```diff
@@ -0,0 +1,148 @@
...
+    output_dir.mkdir(exist_ok=True)
+
+    for i in range(1, num_files + 1):
+        generated_code = generator.generate_statements(num_statements)
+        generator.reset()
```
tianhan0, line 30: Could we swap these two lines? I think it is cleaner to always reset before generating code.

scripts/pysa_fuzzer/run.py
```diff
@@ -0,0 +1,148 @@
...
+
+def run_pyre():
+    logging.info("Please wait a minute or two for Pysa to Run!")
+    run_command(['pyre', '-n', 'analyze'], output_file='analysis_output.tmp')
+
+def find_undetected_files():
+    # Load the number of files from the temp file
+    try:
+        with open('../num_files.tmp', 'r') as tmp_file:
+            num_files = int(tmp_file.read().strip())
```
tianhan0, line 80: An alternative is to directly remember the names of all files. This could be arguably simpler, because now we don't have to re-generate `all_files` (line 99).

scripts/pysa_fuzzer/run.py
```diff
@@ -0,0 +1,148 @@
...
+
+    if args.action == 'all':
+        generate_python_files(args.num_files, args.num_statements)
+        configure_and_analyze()
+        run_pyre()
+        # Save the number of files to a temporary file
+        with open('num_files.tmp', 'w') as tmp_file:
+            tmp_file.write(str(args.num_files))
+    elif args.action == "analyze":
+        find_undetected_files()
```
tianhan0, line 143: What does action `analyze` mean? If it means finding the files where Pysa can't find a flow, maybe a more descriptive name is just `find-undetected-files`? But this is a bit too long -- naming is hard. Feel free to suggest alternatives -- any more descriptive names would be better.

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Generate the non-tainted variable initialization
+        non_tainted_initialization = f'{non_tainted_var} = "temp"'
+
+        # Randomly decide which branch is the valid one
+        valid_branch = random.randint(1, 3)
+
+        if valid_branch == 1:
+            conditions = [
+                (f"if {prev_var} == {prev_var}:", f"{curr_var} = {prev_var} + {non_tainted_var}"),
```
tianhan0, line 210: How about using something like `random.randint(1, 100) == N` in the conditions? Here `N` can be replaced by any integers. This seems a more intuitive simulation of non-deterministically taking a branch.

Same for below.

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,299 @@
...
+            f"    {textwrap.indent(loop_body, '    ')}"
+        )
+
+    def generate_if_else_elif(self) -> str:
+        prev_var = self.get_last_variable()
+        non_tainted_var = self.generate_new_variable()
+        curr_var = self.generate_new_variable()
+
+        # Generate the non-tainted variable initialization
+        non_tainted_initialization = f'{non_tainted_var} = "temp"'
```
tianhan0, line 203: Maybe
```
non_tainted_initialization = f'{non_tainted_var} = "not_tainted"'
```
So that the generated code is easier to read.

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Randomly decide which branch is the valid one
+        valid_branch = random.randint(1, 3)
+
+        if valid_branch == 1:
+            conditions = [
+                (f"if {prev_var} == {prev_var}:", f"{curr_var} = {prev_var} + {non_tainted_var}"),
+                (f"elif {prev_var} == 'invalid':", f"{curr_var} = {non_tainted_var}"),
+                (f"else:", f"{curr_var} = {non_tainted_var}")
```
tianhan0, line 212: An alternative is 
```
(f"else:", f"{curr_var} = 1")
```
since numbers are always not tainted. But it's up to you.

tianhan0 (changes requested):
Thanks for the quick turnaround! Overall looks good to me. But I also left a few comments.

scripts/pysa_fuzzer/run.py
esohel30: I believe just find-undetected suffices.

scripts/pysa_fuzzer/code_generator.py
esohel30: I am a little confused on this. Isn't there a chance that the flow will be lost in this case ?

tianhan0: No that won't be the case. Pysa over-approximates the behavior of the program. In this case, Pysa would pretend that all branches are executed, regardless of what the conditions are. Hence, as long as one of the branches lets the flow go through, the flow would be preserved.

To clarify, my suggestion is more about "aesthetics" -- your code also works perfectly for Pysa, but it is a little less intuitive for a human to understand IMO.

scripts/pysa_fuzzer/code_generator2.py
```diff
@@ -0,0 +1,123 @@
...
+        new_var = self.generate_new_variable()
+
+        current_function_sink = self.generate_new_function()
+        temp_sink = self.last_sink
+        self.last_sink = current_function_sink
+
+        self.source_statements.append(f"def {current_function_source}():\n{indent_space}random_number = random.randint(1,3)\n{indent_space}{new_var} = {temp_source}\n{indent_space}if random_number == 1:\n{indent_space*2}return {new_var}\n{indent_space}else:\n{indent_space*2}return {current_function_source}()")
+        self.sink_statements.append(f"def {current_function_sink}(x):\n{indent_space}random_number = random.randint(1,3)\n{indent_space}if random_number == 1:\n{indent_space*2}return {temp_sink}(x)\n{indent_space}else:\n{indent_space*2}return {current_function_sink}(x)")
+
+    def add_variable(self) -> None:
```
tianhan0, line 54: Maybe rename this into `add_source_variable`, since we only propagate down the source taint?

We can also have an `add_sink_variable`, to propagate up the sink taint, but this won't work for now since your assumption is that `last_sink` is a function name, instead of an expression. By contrast, your `last_source` is an expression, which is good. See the definition of an expression [here](https://en.wikipedia.org/wiki/Expression_(computer_science)).

I would suggest to let `last_sink` always be an expression with a "hole", where the hole will eventually be filled with `last_source`.

For example, assume `last_sink` is `print(#hole)`, where `#hole` can be eventually replaced with `last_source`.

Then, to add a function, we can generate a sink statement
```
def f(x):
  print(x)  # Note here we replace `#hole` with an arbitrary variable
```
and update `last_sink` to be `f(#hole)`.

Alternatively, we can add a sink variable. That is, we can generate a sink statement
```
x = #hole
print(x)  # Note here we also replace `#hole` with an arbitrary variable
```
and update `last_sink` to be `#hole`.

Now, in function `generate`, we would
1. concatenate the source and sink statements, and
2. replace `#hole` in the sink statements (note that there should always exist exactly one `#hole` in all sink statements) with `last_source`.

In this way, we no longer need `generate_sink`, because its functionality is achieved by 2 above.

tianhan0 (changes requested):

scripts/pysa_fuzzer/code_generator2.py
```diff
@@ -0,0 +1,194 @@
...
+
+class CodeGenerator:
+    def __init__(self) -> None:
+        self.variables = self.generate_variable_names()
+        self.current_var = 1
+        self.current_function_number = 1
+        self.source_statements = ['a = input()']
+        self.last_source = "a"
+        self.sink_statements = []
+        self.last_sink = "print"
```
arthaud, line 13: It would be more interesting to start with:
```suggestion
        self.source_statements = []
        self.last_source = "input()"
        self.sink_statements = []
        self.last_sink = "print"
```
In a way, storing the taint in a temporary variable is already a mutation.

tianhan0 (changes requested):
I like the mutations that you wrote. But it feels like we could be more systematic / structured when designing mutations. Currently, many of your mutations are not "ad-hoc" combinations, but there is no principle behind it.

The syntax of a program can usually be described with Backus–Naur form. See https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form. Hence our mutations can also mimic how a program's syntax is defined (since we are generating programs).

For example, below I define a program
```
x ∈ Variables
e ::= True | False | 0 | 1 | 2 | ... | x | x + y | ...
s ::= x = e | ... | if e then s1 else s2 | while e s | s1; s2
```
Here `x` means a variable. `e` is an expression. `s` is a statement. `s1; s2` means sequential statements -- `s1` followed by `s2`.
Note that the above grammar is *not* in Python syntax, because here I'm just trying to give an idea.

Now, each mutation could be simply be one construct under `s`, such as
- an assignment `x = e` (which is `source_mutation_2`), or
- an if-then-else statement `if e then s1 else s2`, which is a more generic version of `source_mutation_3`, because here `s1` and `s2` can be further instantiated as other statements (hence your implementation should recursively generate further statements for `s1` and `s2`), but for example in `source_mutation_3` you instantiate `s1` to be a `print`.

We need to ensure that, each construct propagates the taint. For example, if a source mutation is an assignment `x = e`, then we need to make sure `x` always has the source taint.

Overall, some of your mutations should be described by a compound statement, such as an if-then-else statement `if e then s1 else s2` or a sequential statement `s1; s2`. Other mutations should be described by an atomic statement, such as assignment `x = e`.
Then, when choosing a mutation, we are simply choosing some compound statement or some atomic statement. If we choose a compound statement, we should further specify "how far" should we go. For example, for an `if e then s1 else s2`, you can instantiate `s1` with arbitrarily long statements, but we do need to stop somewhere.

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,324 @@
...
+    def generate_for_loop(self, depth: int = 1) -> str:
+        prev_var = self.get_last_variable()
+        curr_var = self.generate_new_variable()
+        loop_body = f"{curr_var} += {prev_var}"
+
+        if depth > 1:
+            nested_loop = self.generate_for_loop(depth - 1)
+            loop_body = f"{nested_loop}\n{textwrap.indent(loop_body, '    ')}"
+
+        return f"{curr_var} = ''\nfor _ in range({random.randint(2, 5)}):\n{textwrap.indent(loop_body, '    ')}"
```
arthaud, line 63: here and other places: could we do something like this (see below) so it's easier to read?
```suggestion
        return (
          f"{curr_var} = ''\n"
          + f"for _ in range({random.randint(2, 5)}):\n"
          + f"{textwrap.indent(loop_body, '    ')}")

```

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,324 @@
...
+            conditions = [
+                (f"if {prev_var} == {prev_var}:", f"{curr_var} = {prev_var} + {non_tainted_var}"),
+                (f"elif {prev_var} == 'invalid':", f"{curr_var} = {non_tainted_var}"),
+                (f"else:", f"{curr_var} = {non_tainted_var}")
+            ]
+        elif valid_branch == 2:
+            conditions = [
+                (f"if {prev_var} == 'invalid':", f"{curr_var} = {non_tainted_var}"),
+                (f"elif {prev_var} == {prev_var}:", f"{curr_var} = {prev_var} + {non_tainted_var}"),
+                (f"else:", f"{curr_var} = {non_tainted_var}")
```
arthaud, line 218: We don't really need to use the variable in the condition.
What about using something like:
```suggestion
                (f"if random.random() < 0.5:", f"{curr_var} = {non_tainted_var}"),
                (f"elif random.random() < 0.5:", f"{curr_var} = {prev_var} + {non_tainted_var}"),
                (f"else:", f"{curr_var} = {non_tainted_var}")

```

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,324 @@
...
+            ]
+
+        if_else_elif_statements = "\n".join([f"{condition}\n{textwrap.indent(action, '    ')}" for condition, action in conditions])
+        return f"{non_tainted_initialization}\n{if_else_elif_statements}"
+
+    def generate_nested_loops(self) -> str:
+        prev_var = self.get_last_variable()
+        curr_var = self.generate_new_variable()
+        loop_body = f"{curr_var} += {prev_var}"
+        nested_loop = (
```
arthaud, line 234: nitpick: use return directly
```suggestion
        return (
```

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,324 @@
...
+        if number_statements < 2:
+            raise ValueError("number_statements should be at least 2 to include source and sink functions.")
+        
+        function_generators = [
+            lambda: self.generate_addition(random.randint(1, 3)),
+            lambda: self.generate_for_loop(random.randint(1, 3)),
+            lambda: self.generate_while_loop(random.randint(1, 3)),
+            lambda: self.generate_list(random.randint(1, 3)),
+            lambda: self.generate_dictionary(random.randint(1, 3)),
+            lambda: self.generate_function_chain(random.randint(1, 3)),
```
arthaud, line 291: we know that this one is not handled correctly by pysa, right?
Could we add an option that disables the patterns that are known false negatives?

scripts/pysa_fuzzer/code_generator.py
```diff
@@ -0,0 +1,324 @@
...
+        return full_code
+
+
+
+generator = CodeGenerator()
+print(generator.generate_source())
+print(generator.generate_for_loop_with_if())
+print(generator.generate_for_loop_with_if())
+print(generator.generate_for_loop_with_if())
+print(generator.generate_sink())
```
arthaud, line 323: can we remove those, this is only used for testing?

scripts/pysa_fuzzer/example.py
```diff
@@ -0,0 +1,445 @@
+
```
arthaud, line 1: This is useful to see the output, but we probably don't want to merge this. Can you remove it in the final PR?

scripts/pysa_fuzzer/run.py
```diff
@@ -0,0 +1,150 @@
...
+
+def generate_python_files(num_files, num_statements):
+    generator = CodeGenerator()
+    output_dir = Path('generated_files')
+
+    output_dir.mkdir(exist_ok=True)
+
+    filenames = []
+    for i in range(1, num_files + 1):
+        generator.reset()
```
arthaud, line 29: just create a new generator each iteration?
```suggestion
        generator = CodeGenerator()

```

scripts/pysa_fuzzer/run.py
```diff
@@ -0,0 +1,150 @@
...
+            file.write(generated_code)
+        filenames.append(str(filename))
+        logging.info(f"Generated {filename}")
+
+    # Save the filenames to a temporary file
+    with open('filenames.tmp', 'w') as tmp_file:
+        json.dump(filenames, tmp_file)
+
+def configure_and_analyze():
+    with open('.pyre_configuration', 'w') as config_file:
```
arthaud, line 42: can we actually create all the generated files in a directory `generated/`? 
This will also make the implementation of `clean_up` easy, we can use `shutil.rmtree` to delete the directory.

scripts/pysa_fuzzer/run2.py
```diff
@@ -0,0 +1,187 @@
...
+
+def apply_mutation(generator, mutation_number):
+    if 1 <= mutation_number <= 24:
+        # Apply source mutation
+        mutation_method = getattr(generator, f"source_mutation_{mutation_number}")
+        mutation_method()
+    elif 25 <= mutation_number <= 48:
+        # Apply sink mutation
+        mutation_method = getattr(generator, f"sink_mutation_{mutation_number - 24}")
+        mutation_method()
```
arthaud, line 33: could we move this in the CodeGenerator? Otherwise we are leaking some implementation details of the code generator. I would expect the API to be `CodeGenerator.apply_mutation(random_integer)`.

And since we are doing some introspection uglyness, we can actually find automatically all mutations so we don't have to hardcode the 24 and 48:
`source_mutations = {int(name[len('source_mutation_'):]) for name in dir(CodeGenerator) if name.startswith('source_mutation_')}`

scripts/pysa_fuzzer/run2.py
```diff
@@ -0,0 +1,187 @@
...
+    try:
+        result = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True)
+        if output_file:
+            with open(output_file, 'w') as file:
+                file.write(result.stdout)
+    except subprocess.CalledProcessError as e:
+        logging.error(f"Command '{command}' failed with error: {e.stderr}")
+
+def generate_random_seed(num_files, num_mutations=48):
+    return [random.randint(1, num_mutations) for _ in range(num_files)]
```
arthaud, line 23: I wouldn't call this "random_seed".
The seed is the initial number that is used to initialize the random number generator. You aren't setting the seed anywhere so it's generator by the OS.  You can pick a specific seed with `random.seed(42)` for instance.
TL, DR: would call this `generate_mutations_from_seed` and take a `seed` parameter.

scripts/pysa_fuzzer/run2.py
```diff
@@ -0,0 +1,187 @@
...
+
+    filenames = []
+    mutation_mapping = {}
+    for i in range(num_files):
+        mutation_number = seed[i]
+        apply_mutation(generator, mutation_number)
+        generated_code = generator.generate()
+        filename = output_dir / f'test_{i+1}.py'
+        with open(filename, 'w') as file:
+            file.write("import random\n")
```
arthaud, line 49: this should be generated directly by `CodeGenerator.generate()` instead of being added here?

scripts/pysa_fuzzer/run2.py
```diff
@@ -0,0 +1,187 @@
...
+            file.write(generated_code)
+        filenames.append(str(filename))
+        mutation_mapping[filename.name] = mutation_number  # Map the filename to its mutation
```
arthaud, line 52: I don't think we actually use that variable?

scripts/pysa_fuzzer/run2.py
```diff
@@ -0,0 +1,187 @@
+import subprocess
```
arthaud, line 1: could we unify run.py and run2.py? Many of the functions are similar. We just need to have different functions for the code generation. There could be a command line parameter to pick the code generator.

scripts/pysa_fuzzer/code_generator2.py
```diff
@@ -0,0 +1,1878 @@
+import os
```
arthaud, line 1: could we rename code_generator.py and code_generator2.py to be more descriptive?
For instance:
code_generator.py => forward_code_generator.py ("forward" because it generates code from top to bottom)
code_generator2.py => mutation_based_code_generator.py

scripts/pysa_fuzzer/code_generator2.py
```diff
@@ -0,0 +1,1878 @@
...
+        code_lines = self.source_statements + self.sink_statements
+        code_lines.append(self.generate_sink())
+        return '\n'.join(code_lines)
+
+    def source_mutation_1(self) -> None: 
+        indent_space = ' ' * 4
+        current_function = self.generate_new_function()
+        self.source_statements.append(f"""
+def {current_function}(x):
+{indent_space}return x
```
arthaud, line 52: Here and anywhere we use `indent_space`: why not just add the whitespaces directly? That seems more readable.
```suggestion
    return x
```

scripts/pysa_fuzzer/code_generator2.py
```diff
@@ -0,0 +1,1878 @@
...
+            decrypted.extend([square[row_a][(col_a - 1) % 5], square[row_b][(col_b - 1) % 5]])
+        elif col_a == col_b:
+            decrypted.extend([square[(row_a - 1) % 5][col_a], square[(row_b - 1) % 5][col_b]])
+        else:
+            decrypted.extend([square[row_a][col_b], square[row_b][col_a]])
+    return ''.join(decrypted)
+
+y = f40(f39(w, z), z)
+
+print(y)
```
arthaud, line 1878: Remove all this?

arthaud (changes requested):

alexkassil:
Congrats and well done @esohel30 !

arthaud:
Thanks for the hard work! We have finally merged this and found a few false negatives (see upcommit commits).

PR #890, opened 2024-06-26
added show to registry

esohel30:
#869 

This pull request adds pp and show functions to the Registry module to enable pretty-printing of the map contents for easier debugging and logging.

Changes:

Added pp Function:

The pp function formats the map's key-value pairs for pretty-printing.
Utilizes Format.fprintf for formatting and printing the map contents in a readable format.

Added show Function:

The show function converts the map to a string representation using the pp function.

Uses Format.asprintf to generate the string representation of the map.

### DIFF
--- a/source/interprocedural/fixpointAnalysis.ml
+++ b/source/interprocedural/fixpointAnalysis.ml
@@ -163,6 +163,30 @@ module Make (Analysis : ANALYSIS) = struct
 
 
     let get target registry = Target.Map.find_opt registry target
+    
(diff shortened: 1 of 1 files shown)


esohel30:
Hey maxime I am kind of confused about pretty-printing a registry. How do I go about doing it within a test. I understand that I  don't have to make it a test and that I just need to demonstrate that it works by showing some logs/stderr output but I am still confused on how to go about doing that.

arthaud:
Done in https://github.com/facebook/pyre-check/pull/877

PR #893, opened 2024-07-10
Try fixing github build: install only opam with apt

stroxler:
Summary:
The build in github is currently broken due to a dune version mismatch.

I think the problem is that our workflow is incorrectly getting `dune` from
an apt download rather than our opam switch; trying to see what happens if
we only apt install opam.

Differential Revision: D59591828

### DIFF
--- a/.github/workflows/pysa.yml
+++ b/.github/workflows/pysa.yml
@@ -21,23 +21,24 @@ jobs:
       - name: Install Dependencies
         run: |
           pip install -r requirements.txt
-          sudo apt-get install ocaml ocaml-dune
-
-      - name: Setup OCaml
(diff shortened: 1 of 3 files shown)


stroxler:
I'm going to leave this up for now... the latest results seem to show that `setup.py`'s PATH is somehow not making it all the way into the makefile logic behind `make dev`.

See these logs:
https://github.com/facebook/pyre-check/actions/runs/9911237956/job/27383440089?pr=893

It's pretty easy to see the path logged out from `setup.py` does not match what the Makefile sees, and this is why the Makefile can't find `dune`.

I'm pretty stumped about why; I know for sure that environment variables set at the top level *do* make it.

PR #895, opened 2024-07-10
Integration test for node false negative pattern 

esohel30:
adding integration test for node false negative pattern.

### DIFF
--- a/source/interprocedural_analyses/taint/test/integration/node_patterns.py
+++ b/source/interprocedural_analyses/taint/test/integration/node_patterns.py
@@ -1,0 +1,17 @@
+from builtins import _test_sink, _test_source
+
+class Node:
+    def __init__(self, val=None):
+        self.val = val
+        self.next = None
+
+def linked_list_pattern(x):
+    b = Node(x)
+    c = Node()
+    d = Node()
+
+    d.next = c 
+    c.next = b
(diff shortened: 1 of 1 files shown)


tianhan0 (changes requested):
Thanks for adding the test!

Could you also include the expected results for this test (e.g., the `*.models` files)? Otherwise the CI tests would fail.
You can run Pysa on all integration tests by executing `OUNIT_SHARDS=1 dune exec interprocedural_analyses/taint/test/integrationTest.exe` under `pyre/source`. This would
1. either create new expected results for new tests (such as this PR), because there is no existing expected results.
2. or create the actual results, when they differ from the existing expected results.

Additionally, could we
1. rename this test to be more descriptive, such as `heap_alias.py`? We try to let the name of each integration test be as descriptive as possible.
2. add a few comments in the .py file to explain why this is a false negative. You could paraphrase Maxime's explanation (in the chat).

kinto0:
cc @tianhan0 for updated review

tianhan0:
> cc @tianhan0 for updated review

@kinto0 I don't think this PR has been updated in the past month. But we will keep an eye on this -- if the author (who is an MLH fellow) does not update this in another month or so, we will abandon this PR and recreate it internally.

PR #945, opened 2024-11-20
Bump black and ufmt versions

stroxler:
Summary:
I'm seeing mismatches between what our IDEs are doing and what
github CI says it wants. My current guess is that this is due
to stale versions in github CI.

Differential Revision: D66269903


stroxler:
Unfortunately this did not help, unsure why usort is unhappy here

PR #953, opened 2024-12-12
Use pyre2/pyre2 as working directory for pyre2 github workflow

rchen152:
Summary:
The workflow is currently failing with "error: could not find `Cargo.toml` in `/home/runner/work/pyre-check/pyre-check` or any parent directory" (example: https://github.com/facebook/pyre-check/actions/runs/12285231821).

working-directory stuff copied from https://stackoverflow.com/a/63122434.

Differential Revision: D67113206


ndmitchell:
Already fixed on master

PR #998, opened 2025-07-07
fix: use enum value for Click version option flag_value

er-nabin-bhusal:
**Pre-submission checklist**
- [ ] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [ ] `pre-commit run`

## Summary
The expectation in the click.option is a literal but it was getting an enum type, thus now we are passing value of enum. 

I was trying to setup pyre in my django project and i got the error as in sc.
image

### DIFF
--- a/client/pyre.py
+++ b/client/pyre.py
@@ -186,8 +186,8 @@ def _create_and_check_configuration(
     "--version",
     is_flag=False,
     type=click.Choice([kind.value for kind in command_arguments.VersionKind]),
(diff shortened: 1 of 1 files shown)


er-nabin-bhusal:
image
Updating the version seems to fix the issue but I don't want to suggest it yet, because I don't the implication of this outside of this.

PR #1050, opened 2026-05-28
[pysa] Update expected out of deliberately vulnerable flask app

arthaud:
It looks like the false negative regarding open redirect is now fixed.


stroxler (approved):
Review automatically exported from Phabricator review in Meta.

